PR #1, opened 2019-09-23
TIG-1951 Create selected-tests service and add a health API endpoint

lydiastepanek:
This PR sets up a /health endpoint for the selected tests service. It can be accessed locally or in kanopy at the following links, respectively:
* http://localhost:8080/health
* https://selected-tests.server-tig.prod.corp.mongodb.com/health

This PR sets up the evergreen and kanopy integrations for this project.

I am planning on creating a wiki page with the following information about the design decisions made in this PR (subject to change as this PR goes through review -- I will update this list as comments come in):
* This cookie cutter python template was used for the project layout: https://github.com/dbradf/py-cli-template
* Flask + Gunicorn for the web server
* Python 3 
* flake8 for PEP8 compliance checking
* black for formatting
* Pytest for testing


evergreen.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: unit_tests
+  commands:
+  - command: shell.exec
+    params:
+      working_dir: src
+      script: |
+        set -o errexit
+        . venv/bin/activate
+
+        pytest --flake8 --ignore=venv
```
lydiastepanek: flake8 is checking dependency packages here (the evergreen logs [show](https://evergreen.mongodb.com/task_log_raw/selected_tests_ubuntu1604_unit_tests_patch_df063b57db3321f8255f98fc06e2e5f4fb998783_5d8bd9c02a60ed71a4405ab1_19_09_25_21_18_57/0?type=T#L88) a flake8 warning for the jinja2 package) and I'm not sure why

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ubuntu:18.04
```
zamj: It'll probably speed up our builds considerably and make them a bit more predictable if we moved to using a different docker base image, such as the official python3 base image. As a nice benefit, it'll let us simplify the logic here to something like 

```
FROM python:3
ADD . /selected-tests
WORKDIR /selected-tests 
RUN pip install -r requirements.txt
RUN pip install -e .

CMD ["scripts/server.sh"]
```

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    && apt-get install -y libffi-dev \
+    && apt-get install -y libssl-dev \
+    && apt-get install -y python3-pip \
+    && apt-get install -y git
+RUN pip3 install --upgrade pip
+
+ADD . /selected-tests
+WORKDIR /selected-tests
+RUN pip3 install -r requirements.txt
+RUN pip3 install -e .
```
zamj: Not very familiar with pip's intricacies, what does this line do? Do we need it?

scripts/server.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+export LC_ALL=C.UTF-8
+export LANG=C.UTF-8
+
+gunicorn --bind 0.0.0.0:8080 --workers 3 selectedtests.app.wsgi
```
zamj: [nit] It might make these settings a little more manageable in the long run if we put these in a separate config file and passed that as an argument instead. 

http://docs.gunicorn.org/en/stable/configure.html#configuration-file

tests/selectedtests/test_app.py
```diff
@@ -0,0 +1,22 @@
...
+from selectedtests.app import app as under_test
+from flask import testing
+
+
+@pytest.fixture()
+def app_client():
+    """
+    Client for the flask web with mocked endpoints.
+    """
+    client = under_test.create_app().test_client()
```
ancostas: You mentioned not wanting to share app state across tests. You can avoid changing the app's architecture (`create_app` function wrapping everything) by changing this to:
```
@pytest.fixture()
def app_client():
    """
    Client for the flask web with mocked endpoints.
    """
    from selectedtests.app import app as under_test
    client = under_test.application.test_client()
    return client
```

.python-version
```diff
@@ -0,0 +1 @@
+selected-tests-3.7.3
```
dbradf: This file should be in .gitignore.

Dockerfile
dbradf: This installs the package in the current directory in editable mode. We need this line, but we probably don't need the `-e` flag.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    && apt-get install -y python3-dev \
+    && apt-get install -y libffi-dev \
+    && apt-get install -y libssl-dev \
+    && apt-get install -y python3-pip \
+    && apt-get install -y git
+RUN pip3 install --upgrade pip
+
+ADD . /selected-tests
+WORKDIR /selected-tests
+RUN pip3 install -r requirements.txt
```
dbradf: I don't think we need to install `requirements.txt`. It should only contain things we need for development.

README.md
```diff
@@ -1 +1,44 @@
...
+```
+$ black src
+```
+
+See the black [documentation](https://github.com/psf/black#editor-integration) for details on how
+to configure your editor to automatically format your code.
+
+## Deploy
+
+Deployment is done via helm to [canopy](https://github.com/10gen/kanopy-docs#index). The project
```
dbradf: [nit] s/canopy/kanopy/

requirements.txt
```diff
@@ -0,0 +1,5 @@
+black==19.3b0
+flask==1.0.3
```
dbradf: flask and gunicorn should probably be dependencies in setup.py.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 100
```
dbradf: We might want to check if we can put this in the pyproject.toml.

setup.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import print_function
```
dbradf: I don't think we need this, since we will only support python 3.

setup.py
```diff
@@ -0,0 +1,44 @@
...
+        'Operating System :: Unix',
+        'Operating System :: POSIX',
+        'Operating System :: Microsoft :: Windows',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: Implementation :: CPython',
+        'Programming Language :: Python :: Implementation :: PyPy',
+    ],
+    install_requires=[],
+    entry_points='',
```
dbradf: It might be nice to have an entry_point to start the webapp.

src/selectedtests/app/app.py
```diff
@@ -0,0 +1,34 @@
...
+    :return: Instance of flask application.
+    """
+    app = Flask(__name__)
+
+    @app.route("/health")
+    def health():
+        """
+       Get information about whether service is running
+       """
+        return "ok"
```
dbradf: We might consider returning json here. That would make it easy to expand in the future.

Dockerfile
dbradf: See https://pythonspeed.com/articles/base-image-python-docker-images/

I don't think we should use just `python:3`, we should explicitly pick which distro to use. I have had problems with alpine in the past and would prefer not to use it.

README.md
```diff
@@ -1 +1,44 @@
...
+$ pytest --cov=src --cov-report=html
+```
+
+## Style
+
+This project is formatting with [black](https://github.com/psf/black). To autoformat your code, you
+can use the black command line tool:
+
+```
+$ black src
```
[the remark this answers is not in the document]
dbradf: In other projects we aren't formatting the tests. But if we wanted to we could.

evergreen.yml
dbradf: It has to do with the way the flake8 analysis happens. Since jinja2 packages get imported into other files, they get analyzed. 

There is conversation about it [here](https://github.com/pallets/jinja/pull/867). It sounds like updating the version of flask we are using might help (if it has been released).

Dockerfile
dbradf: Also, just to note, every other project we have is using the ubuntu image. So using a different image could lead to this project behaving differently. If we used the python:3 debian image, they should be similar, but there still could be differences.

dbradf:
> In regards to the directory names, I think it makes sense here to have some sort of separator between words for readability's sake, especially as one of the words is "tests". It looks like the suggested separator is an `_` between words.
> 
> https://www.python.org/dev/peps/pep-0008/#package-and-module-names

The link you posted says the opposite of what you are requesting:

> Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

Dockerfile
zamj: I don't really see a risk in the distro being different across repos. Everything we're doing in them is just using the python commands and runtime so it's rather unlikely to be different across debian and ubuntu distros.

That's a fair point about alpine, it can be rather annoying at times. And we can pin it further to a specific minor or even patch version if people want.

```3.7.4-buster, 3.7-buster, 3-buster, buster```
```3.7.4-stretch, 3.7-stretch, 3-stretch, stretch```

All the above are available in docker hub. 

If moving to something besides ubuntu isn't favorable though, another option is to build an ubuntu python base image and push it to a registry and use that.

**Comment by zamj** [comment]
> > In regards to the directory names, I think it makes sense here to have some sort of separator between words for readability's sake, especially as one of the words is "tests". It looks like the suggested separator is an `_` between words.
> > https://www.python.org/dev/peps/pep-0008/#package-and-module-names
> 
> The link you posted says the opposite of what you are requesting:
> 
> > Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

Whoops got confused by the sentence before about modules. Shows me to read everything

File: Dockerfile (hunk 2 above)
Comment (lydiastepanek) [reply Dockerfile]: Sounds like we want to stick to ubuntu for consistency, so I'm going to leave this as-is for now. I don't think there's a huge benefit from building a ubuntu python base image (given this Dockerfile is relatively small as-is) but am happy to discuss as a separate ticket.

Comment (zamj) [reply Dockerfile]: The main benefit to building an ubuntu image is the guarantee of the same runtime environment every time we build a new version off of that base. Currently, we're installing the same packages into a base ubuntu version every build which theoretically should give the same/a similar enough result but doesn't give a guarantee of it. 

The other big benefit here is build speed as rather than having to install the same packages across builds they'd already be installed into the base and all the logic in the dockerfile would have to do with how to run our code rather than a runtime. 

If the decision is to table it for now, that's fine.

File: scripts/server.sh (hunk 4 above)
Comment (lydiastepanek) [reply scripts/server.sh]: There are only two settings right now, but if we add more I am happy to break it out into a config file

File: README.md [hunk 15]
```diff
@@ -1 +1,44 @@
-# selected-tests
+# Selected Tests
+
+This service is used to predict which tests need to run based on code changes.
+
+# Set up environment
```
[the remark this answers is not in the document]
lydiastepanek: I considered adding this, but we don't mention virtual environments in our other python READMEs, I think virtual environments are assumed

README.md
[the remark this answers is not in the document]
lydiastepanek: Sure 👍

Dockerfile
dbradf: I think we table it for now.

We can have a separate discussion for what we want to do as a team and if we want switch over the Dockerfiles in all our projects to whatever we decide on.

README.md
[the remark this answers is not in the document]
dbradf: Using a virtualenv isn't a requirement. It is a really good idea to keep dependencies separate, but if a developer really want to manage environments in a different way they could. We should probably have something on the wiki that talks about using virtualenv because it would be the same for all projects.

requirements.txt
ancostas: Do we really need gunicorn to run this? Seems like that is a deployment related detail that can go in the deployment pipeline instead?

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    && apt-get update \
+    && apt-get install -y curl gcc g++ \
+    && apt-get install -y python3-dev \
+    && apt-get install -y libffi-dev \
+    && apt-get install -y libssl-dev \
+    && apt-get install -y python3-pip \
+    && apt-get install -y git
+RUN pip3 install --upgrade pip
+
+ADD . /selected-tests
```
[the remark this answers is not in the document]
lydiastepanek: see https://github.com/mongodb/selected-tests/pull/1#discussion_r328730679 for related conversation

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2019 lydia
```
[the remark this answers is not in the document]
lydiastepanek: nice catch

setup.py
```diff
@@ -0,0 +1,44 @@
...
+
+from setuptools import find_packages
+from setuptools import setup
+
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
+
+setup(
+    name='selectedtests',
```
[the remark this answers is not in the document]
lydiastepanek: What do you recommend? I'd also probably have to rename it in the [technical design](https://docs.google.com/document/d/1azKrJr3babowhr6M8vs9cCTQOmLdUeyqmyVKp_ENSOw/edit), but I am open to changing it.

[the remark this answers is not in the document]
ancostas: I thin David and Jeff have more to say about naming than I do. I'll drop my request since I've not come up with a good name, and snake case is apparently against python standards.

ancostas:
@dbradf @zamj I've assigned you two to this so you can monitor the PR under `Pull Requests > Assigned` rather than having to watch your emails. The PR disappears from 1Pull Requests > Review Requested1 after the initial round of reviews.

src/selectedtests/app/wsgi.py
```diff
@@ -0,0 +1,3 @@
+from selectedtests.app.app import main
```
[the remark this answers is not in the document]
lydiastepanek: I moved `wsgi.py` out of `src/` 👍 

I don't want to simplify app.py the way you suggest, it is easier to test as-is, plus if we want to avoid dependency injection later on, then `app` can be passed as an argument to `create_app`, rather than being defined on the file level.

Added an entry point 👍

tests/selectedtests/test_app.py
lydiastepanek: I don't want to have to re-import `selectedtests.app` module every time the `app_client` fixture is called.

requirements.txt
lydiastepanek: Yes, we need gunicorn in setup.py since the Dockerfile runs `pip3 install .` and needs gunicorn to be installed in order to run.

src/selectedtests/app/wsgi.py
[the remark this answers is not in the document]
lydiastepanek: Actually, since `wsgi.py` is used by gunicorn, and gunicorn is required in setup.py, I think it's fair to include `wsgi.py` in `src/` so I'm adding it back.

setup.cfg
lydiastepanek: 👍 Updated flake8 commands to specify pyproject.toml as config file

lydiastepanek: Actually, flake8 wouldn't recognize the config gile, so I switched to having black specify setup.cfg for the config file

evergreen.yml
lydiastepanek: I updated flask but it didn't fix it, so I'll leave it as-is for now

requirements.txt
```diff
@@ -0,0 +1,5 @@
+black==19.3b0
+flask==1.0.4
```
dbradf: [nit] Since 'flask' is specified in the setup.py file dependencies, it shouldn't be needed here.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+WORKDIR /selected-tests
+RUN pip3 install .
+RUN pip install gunicorn==19.9.0
```
zamj: Is this pip intentionally different from the one above?

setup.py
```diff
@@ -0,0 +1,47 @@
...
+# -*- encoding: utf-8 -*-
+from glob import glob
+from os.path import basename
+from os.path import splitext
+
+from setuptools import find_packages
+from setuptools import setup
+
+
+with open("README.md", "r") as fh:
```
zamj: You'll want to specify what kind of encoding you're using if this is going to run in kanopy. Pip will get confused otherwise.

Something like this if it's actually utf-8.
```
with open("README.md", "r", encoding='utf-8') as fh:
```

dbradf: This shouldn't matter. This code is reading the readme file to create a description for the packaging information. So it is really only important if we are building the package in order to upload it to PyPi. Since we aren't planning to upload this project to PyPi and we wouldn't upload the package from Kanopy if we were, we don't really need to specify the encoding.

Dockerfile
lydiastepanek: Yes, gunicorn is not a requirement in setup.py (the requirements in setup.py are required for the package to run and will get installed with `pip3 install .`), but gunicorn is needed for the Docker container to run (since it gets called in `scripts/server.sh`), so I needed to install it in the Docker image.

setup.py
zamj: I'm fairly certain an error will still be thrown when the module is built locally in the docker container. But I took another look and the encoding is being specified in the web_server script anyways so this is probably fine. Going to resolve.

Dockerfile
lydiastepanek: Oh, you meant pip vs. pip3 :). Switching to pip3.

PR #2, opened 2019-10-01
TIG-1956: Add openapi support

zamj:


requirements.txt
```diff
@@ -1,4 +1,4 @@
 black==19.3b0
 gunicorn==19.9.0
 pytest-flake8==1.0.4
-pytest==4.2.1
+pytest==5.2.0
```
zamj: Was facing an issue where pytest would fail to run. Looks like it's linked to this issue: https://github.com/pytest-dev/pytest/issues/3280

The new version is the latest released.

src/selectedtests/app/controllers/health_controller.py
```diff
@@ -0,0 +1,18 @@
+from flask import jsonify
+from flask_restplus import Api, Resource, fields
+
+
+def add_health_endpoints(api: Api):
```
dbradf: Could you add a docstring to this function?

src/selectedtests/app/controllers/health_controller.py
```diff
@@ -0,0 +1,18 @@
...
+
+    ns = api.namespace("health", description="Health check endpoint")
+
+    health_model = ns.model("HealthCheckResponse", {"online": fields.Boolean})
+
+    @ns.route("")
+    class Health(Resource):
```
dbradf: Could you add a docstring to this class?

zamj: Done

src/selectedtests/app/controllers/health_controller.py
zamj: Done

PR #3, opened 2019-10-01
TIG-1956: Fix cors swagger issue

zamj:
The swagger UI is not being served due to a cors issue. Looks to be the issue found here:
https://github.com/noirbizarre/flask-restplus/issues/701

This pr has the code in the link that should fix it.


zamj:
@lydiastepanek As discussed, I've updated to the latest version of flask and also added some very basic sanity tests for the swagger endpoints.

PR #7, opened 2019-10-03
Pydocstyle

zamj:


pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 100
+target-version = ['py37']
```
zamj: We needed to move this out of setup.cfg due to issues with the config for pydocstyle. Read more [here](https://github.com/psf/black/issues/900)

README.md
```diff
@@ -43,12 +43,22 @@ This project is formatting with [black](https://github.com/psf/black). To autofo
 can use the black command line tool:
 
 ```
-$ black src tests --config setup.cfg
+$ black src tests --config pyproject.toml
```
dbradf: [nit] If this is being run the root directory, I think black will automatically detect and use the pyproject.toml file, i.e. I don't think you need to specify `--config`.

zamj: Makes sense - I've updated this accordingly

PR #10, opened 2019-10-22
TIG-1992 CLI interface to create test mappings

lydiastepanek:
https://jira.mongodb.org/browse/TIG-1992


src/selectedtests/task_mappings/mappings.py
```diff
@@ -35,7 +35,6 @@ def create_task_mappings(
         evergreen_project: str,
         start_date: datetime,
         end_date: datetime,
-        org_name: str,
```
lydiastepanek: I removed org_name here because I realized we can get it from the /projects endpoint. See this ticket for more info: https://jira.mongodb.org/browse/EVG-6878.

src/selectedtests/test_mappings/cli.py
```diff
@@ -0,0 +1,191 @@
...
+                end_date,
+                evergreen_project,
+                module_info["module_branch"],
+            )
+            module_test_mappings_list = module_test_mappings.get_mappings()
+            project_test_mappings_list.append(module_test_mappings_list)
+
+    json_dump = json.dumps(project_test_mappings_list, indent=4)
+
+    if output_file is not None and output_file != "":
```
dbradf: I think you can shorten this to just:
```
if output_file:
```

src/selectedtests/test_mappings/cli.py
```diff
@@ -0,0 +1,191 @@
...
+        f = open(output_file, "a")
```
dbradf: [nit] In python it is usually recommend to use `with` statements for opening files. This will automatically close the file when exiting the `with` so you don't need to worry about calling `close`.

```
with open(output_file, "a") as f:
    f.write(json_dump)
````

File: src/selectedtests/test_mappings/mappings.py [hunk 4]
```diff
@@ -0,0 +1,125 @@
...
+import os.path
+
+from datetime import datetime
+from collections import defaultdict
+from structlog.stdlib import LoggerFactory
+from selectedtests.git_helper import modified_files_for_commit
+from re import Pattern
+from git import Repo
+
+structlog.configure(logger_factory=LoggerFactory())
```
dbradf: [nit] I don't think we want to configure the logger here.

src/selectedtests/test_mappings/mappings.py
```diff
@@ -0,0 +1,125 @@
...
+        file_count_map: defaultdict,
+        project: str,
+        repo_name: str,
+        branch: str,
+    ):
+        """
+        Create a TestMappings object.
+
+        :param file_intersection: Map of how files intersect.
+        :param file_count_map: Map of how many times files where seen.
```
dbradf: Could you add `:param` lines for all the parameters?

src/selectedtests/test_mappings/mappings.py
```diff
@@ -0,0 +1,125 @@
...
+        :param start_date: The date at which to start analyzing commits of the project.
+        :param end_date: The date up to which we should analyze commits of the project.
+        :param project: The name of the evergreen project to analyze.
+        :param branch: The branch of the git repo used for the evergreen project.
+        :return: An instance of the test mappings class
+        """
+        file_intersection = defaultdict(lambda: defaultdict(int))
+        file_count = defaultdict(int)
+
+        LOGGER.debug("searching from", ts=start_date)
```
dbradf: [nit] You could probably combine these log statements into one:
```
LOGGER.debug("searching", start_date=start_date, end_date=end_date)
```

src/selectedtests/test_mappings/mappings.py
```diff
@@ -0,0 +1,125 @@
...
+        :return: Transformed test mappings
+        """
+        if not self._test_mappings:
+            self._test_mappings = self._transform_mappings()
+        return self._test_mappings
+
+    def _transform_mappings(self):
+        test_mappings = []
+        for source_file, test_file_count_dict in self._file_intersection.items():
+            test_files = []
```
dbradf: [nit] I think you can use a list comprehension here:
```
test_files = [{"name": test_file, "test_file_seen_count": test_file_seen_count} for test_file, test_file_seen_count in test_file_count_dict.items()]
```

src/selectedtests/test_mappings/cli.py
```diff
@@ -0,0 +1,191 @@
...
+    :param evg_api: An instance of the evg_api client
+    :param project: The name of the evergreen project to analyze.
+    :param module_name: Name of the module to analyze
+    :return: Information about the evergreen module
+    """
+    version_iterator = evg_api.versions_by_project(project)
+    recent_version = next(version_iterator)
+    modules = recent_version.get_manifest().modules
+    module = modules.get(module_repo)
+    return {
```
zamj: It might be nice to, instead of transforming the return type here into a dict, return the module directly. It should have a type of ```ManifestModule```. If we then added that as a return type on the function, it would make the returned value and function a bit easier to use and reason about long term. 

As well, if we ever needed any of the other information about the module, such as the url or revision, we wouldn't have to update this function to add that info to the dict.

src/selectedtests/test_mappings/cli.py
```diff
@@ -0,0 +1,191 @@
...
+
+def _get_evg_project(evg_api: CachedEvergreenApi, project: str) -> Dict:
+    """
+    Fetch an Evergreen project's info from the Evergreen API.
+
+    :param evg_api: An instance of the evg_api client
+    :param project: The name of the evergreen project to analyze.
+    :return: evg_api client instance of the project
+    """
+    for evergreen_project in evg_api.all_projects():
```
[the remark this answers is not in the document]
zamj: The api associated with ```evg_api.project_by_id``` has rather strict permissions where unless you're an admin (I believe that's the correct role) on the project in question you're unable to use that endpoint. That makes it a little difficult for us to use if we're allowing people to tell us what projects to analyze unless we created an evg user with admin permissions to every project in advance.

Lydia talks a bit more about that here https://github.com/mongodb/selected-tests/pull/10#discussion_r337755507

src/selectedtests/test_mappings/mappings.py
```diff
@@ -0,0 +1,125 @@
...
+                "Investigating commit",
+                summary=commit.message.splitlines()[0],
+                ts=commit.committed_datetime,
+                id=commit.hexsha,
+            )
+
+            if commit.committed_datetime.timestamp() < start_date.timestamp():
+                break
+
+            if commit.committed_datetime.timestamp() > end_date.timestamp():
```
zamj: [nit] I'm not sure if you need to convert these datetimes to timestamps here. Datetime comparisons should be the same as timestamp comparisons, where the date farthest in the future is the greater. 

I can't link directly to it but it's right below the linked section: https://docs.python.org/3/library/datetime.html#datetime.datetime.fold

Unless you're converting because one datetime object is timezone aware and the other isn't?

src/selectedtests/test_mappings/mappings.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        Get a transformed version of test mappings to the test mapping object.
+
+        :return: Transformed test mappings
+        """
+        if not self._test_mappings:
+            self._test_mappings = self._transform_mappings()
+        return self._test_mappings
+
+    def _transform_mappings(self):
```
zamj: Can you add some tests for this? To make sure that the format that you expect will be what comes back

src/selectedtests/git_helper.py
```diff
@@ -0,0 +1,79 @@
...
+    :param org_name: The org name in github that owns the repo.
+    :return: An Repo instance that further git operations can be done on.
+    """
+    repo_path = os.path.join(temp_dir, repo_name)
+    url = f"{GITHUB_BASE_URL}/{org_name}/{repo_name}.git"
+    repo = Repo.clone_from(url, repo_path, branch=branch)
+    return repo
+
+
+def _paths_for_iter(diff: Diff, iter_type: str):
```
zamj: Can you add some tests for this?

src/selectedtests/task_mappings/mappings.py
zamj: That'll definitely work in that we do need less information from the user now which is good. However, I'm not super in love with having to iterate over all the projects in evergreen to find the info for the project we're analyzing every time we run the script. 

I think this is a good workaround for getting the needed info but that we should keep the request open for evergreen to expose this information on the ```projects/{projectId}``` endpoint. What do others think?

lydiastepanek: If iterating over all the projects had a high performance cost, I would be willing to keep the request open. But since there's a workaround, the `projects/{projectId}` endpoint change doesn't seem super high priority to me.

dbradf: I think there are around around 300 projects, so iterating over it should be ok. I didn't realize the authenticated user had to have permissions for each project to hit that projects endpoint. That is annoying.

src/selectedtests/test_mappings/mappings.py
lydiastepanek: Yeah, I didn't make start_date and end_date timezone aware. I have now updated them to be UTC and have removed the timestamp conversion.

src/selectedtests/test_mappings/mappings.py
lydiastepanek: That is implicitly tested in the `get_mappings` tests

I tend to favor testing private methods through the public methods that call them: https://www.codewithjason.com/testing-private-methods/

src/selectedtests/git_helper.py
lydiastepanek: Added some tests for the public `modified_files_for_commit` method which calls this

src/selectedtests/task_mappings/mappings.py
lydiastepanek: Created this ticket to follow up on this comment outside of this PR: https://jira.mongodb.org/browse/TIG-2091

src/selectedtests/task_mappings/mappings.py
```diff
@@ -142,9 +140,9 @@ def _get_filtered_files(diff: DiffIndex, regex: Pattern) -> List[str]:
     :return: A list of the changed files that matched the given regex pattern.
     """
     re: List[str] = []
-    for file in _get_changed_files(diff):
-        if match(regex, file.b_path):
-            re.append(file.b_path)
+    for file in get_changed_files(diff, LOGGER):
```
zamj: @dbradf I'm not super familiar with logging in python or structlog but it looks like here https://github.com/dbradf/evgflip/blob/master/src/evgflip/find_flips.py that you're getting the logger instance that you initialized when the cli first ran. Is that correct?

If we can do so, it'd be great if we could remove the ```LOGGER``` param here and just get the logger instance directly in ```git_helper.py```. It would cut down on the params needed by the function and make ```get_changed_files``` a bit more reusable.

dbradf: The big benefit of passing a logger this way as opposed to getting a logger instance in `git_helper.py` is that structlog has the concept of [binding loggers](http://www.structlog.org/en/stable/getting-started.html#building-a-context), which allows you to assign variables to the logger that are printed out with all log entries. 

We aren't using it right now, but it could potentially be useful. For example, in this project it might be useful to bind the project_id to the logger up front and then every log entry made with that logger will include the project_id. When looking through logs that have a lot going on (especially if there are multiple threads or multiple processes), that can be extremely valuable.

So, either method works, it just depends on what we want. Since this project is just getting started, it doesn't really matter. It should be easy to change later.

zamj: Sounds good to me

PR #13, opened 2019-10-29
TIG-1981 Test Mappings Create API to enqueue evergreen project

lydiastepanek:
Uses mongodb queue as detailed here: http://learnmongodbthehardway.com/schema/queues/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,12 +37,13 @@ setup(
         'Programming Language :: Python :: Implementation :: PyPy',
     ],
     install_requires=[
-        'boltons==19.1.0',
         'Click==7.0',
-        'evergreen.py==0.6.9',
-        'flask==1.1.1',
-        'flask-restplus==0.13.0',
         'GitPython==3.0.3',
+        'boltons==19.1.0',
+        'dnspython==1.16.0',
+        'evergreen.py==0.6.9',
+        'flask-restplus==0.13.0',
+        'flask==1.1.1',
         'pymongo[tls]==3.8.0',
         'structlog==19.1.0',
     ],
@@ -51,6 +52,7 @@ setup(
             'selected-tests-service = selectedtests.app.app:main',
             'task-mappings = selectedtests.task_mappings.cli:main',
             'test-mappings = selectedtests.test_mappings.cli:main',
+            'init-mongo = selectedtests.datasource.cli:main',
         ],
     }
 )
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
         'flask==1.1.1',
         'flask-restplus==0.13.0',
         'GitPython==3.0.3',
+        'pymongo[tls]==3.8.0',
         'structlog==19.1.0',
     ],
     entry_points={
--- a/src/selectedtests/app/app.py
+++ b/src/selectedtests/app/app.py
@@ -1,10 +1,16 @@
 """Application to serve API of selected-tests service."""
+import os
 
 from flask import Flask
 from werkzeug.middleware.proxy_fix import ProxyFix
+from evergreen.api import RetryingEvergreenApi
 
 from selectedtests.app.swagger_api import Swagger_Api
 from selectedtests.app.controllers.health_controller import add_health_endpoints
+from selectedtests.app.controllers.project_test_mappings_controller import (
+    add_project_test_mappings_endpoints,
+)
+from selectedtests.datasource.mongo_wrapper import MongoWrapper
 
 DEFAULT_PORT = 8080
 
@@ -25,9 +31,23 @@ def create_app() -> Flask:
         doc="/swagger",
     )
 
+    mongo = _get_mongo_wrapper()
+    evg_api = RetryingEvergreenApi.get_api(use_config_file=True)
+
     add_health_endpoints(api)
+    add_project_test_mappings_endpoints(api, mongo, evg_api)
 
     return app
+
+
+def _get_mongo_wrapper() -> MongoWrapper:
+    """
+    Get an instance of the mongo wrapper based on environment variables.
+
+    :return: MongoWrapper instance.
+    """
+    mongo_uri = os.environ.get("SELECTED_TESTS_MONGO_URI")
+    return MongoWrapper.connect(mongo_uri)
 
 
 def main():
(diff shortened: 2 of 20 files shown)


src/selectedtests/app/test_mapping_work_item.py
```diff
@@ -0,0 +1,120 @@
+"""Model of Evergreen TestMapping that needs to be analyzed."""
```
lydiastepanek: This file is following the conventions of https://github.com/10gen/dag-metrics/blob/master/src/dagmetrics/nr_metrics/version_work_item.py

src/selectedtests/app/app.py
```diff
@@ -25,14 +34,30 @@ def create_app() -> Flask:
...
     return app
 
 
+def _get_mongo_wrapper() -> MongoWrapper:
+    """
+    Get an instance of the mongo wrapper based on environment variables.
+
+    :return: MongoWrapper instance.
+    """
+    mongo_uri = os.environ.get("SELECTED_TESTS_MONGO_URI")
```
lydiastepanek: I have added the `SELECTED_TESTS_MONGO_URI` secret to the `selected-tests` namespace in helm, accessible through:
```
helm ksec get selected-tests
```

setup.py
```diff
@@ -43,6 +43,7 @@
         'flask==1.1.1',
         'flask-restplus==0.13.0',
         'GitPython==3.0.3',
+        'pymongo[tls]==3.8.0',
```
dbradf: If you are planning on using a `mongodb+src` url (which I think atlas uses), you'll need to install [`dnspython`](https://github.com/mongodb/mongo-python-driver#dependencies)

src/selectedtests/app/app.py
```diff
@@ -25,14 +34,30 @@ def create_app() -> Flask:
         doc="/swagger",
     )
 
+    # Creating index no-ops if index already exists
+    setup_indexes(mongo.test_mappings_queue())
```
dbradf: We might want to consider creating the indexes as a separate standalone command. The problem with creating them on application startup is that if we ever want to change the indexes, there has to be some coordination between deleting old indexes and restarting the app.

src/selectedtests/app/controllers/project_test_mappings_controller.py
```diff
@@ -0,0 +1,81 @@
...
+
+def add_project_test_mappings_endpoints(api: Api, mongo: MongoWrapper, evg_api: EvergreenApi):
+    """
+    Add to the given app instance the test mapping jobs endpoints of the service.
+
+    :param api: An instance of a Flask Restplus Api that wraps a Flask instance
+    :param mongo: Mongo Wrapper instance
+    :param evg_api: An instance of the evg_api client
+    """
+    global MONGO_WRAPPER
```
dbradf: I'm not a fan of using global variables like this. It makes the code really difficult to refactor (I've recently had to remove a bunch of these from build-baron-tools to be able to extend it). Plus, I don't think you need global variables in this case, you should just be able to use the function parameters.

src/selectedtests/app/test_mapping_work_item.py
```diff
@@ -0,0 +1,120 @@
...
+import structlog
+
+from datetime import datetime
+from pymongo.errors import DuplicateKeyError
+from pymongo import IndexModel, ASCENDING
+
+LOGGER = structlog.get_logger()
+WORK_ITEM_TTL = 14 * 24 * 60 * 60  # 2 Week TTL.
```
dbradf: You can use `timedelta` to make this a little easier to read.

```
from datetime import timedelta

WORK_ITEM_TTL = timedelta(weeks=2).total_seconds()
```

src/selectedtests/app/test_mapping_work_item.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        Add this work item to the Mongo collection.
+
+        If this work item already exists, just update the created_on timestamp.
+
+        :param collection: Mongo collection containing queue.
+        :return: True if item was new record was added to collection.
+        """
+        LOGGER.info(
+            f"Adding new test_mapping work item for {self.project}, {self.source_file_regex}",
```
dbradf: [nit] I tend avoid f-strings with structlog. It is a personal preference, but I find if there is something I want to add as a f-string variable, it might be more useful to add it as a named-parameter. Because of how named parameters show up in the log, it gives a nice anchor to search on if needed.

src/selectedtests/app/project_test_mapping_work_item.py
```diff
@@ -1,25 +1,26 @@
...
 
     :param collection: Collection to add indexes to.
     """
     index = IndexModel([("project", ASCENDING)], unique=True)
     collection.create_indexes([index])
+    LOGGER.info("Adding indexes for collection", collection=collection.name)
 
 
-class TestMappingWorkItem(object):
+class ProjectTestMappingWorkItemm(object):
```
lydiastepanek: pytest thought this was a test file so I added a "Project" prefix to the class...I think this is more accurate anyway

src/selectedtests/app/controllers/project_test_mappings_controller.py
```diff
@@ -0,0 +1,71 @@
...
+    @ns.route("/<project>/test-mappings")
+    @api.param("project", "The evergreen project identifier")
+    class TestMappingsWorkItem(Resource):
+        @ns.response(200, "Success", test_mappings_work_item)
+        @ns.response(404, "Evergreen project not found", test_mappings_work_item)
+        @ns.response(422, "Work item already exists for project", test_mappings_work_item)
+        @ns.expect(test_mappings_work_item, validate=True)
+        def post(self, project):
+            """Enqueue a project test mapping work item."""
+            evergreen_project = get_evg_project(evg_api, project)
```
zamj: [nit] It may be helpful to do some additional validation here around the modules - that if a ```module``` field is passed in, ```module_source_file_regex``` and ```module_test_file_regex``` fields are required as well like [here](https://github.com/mongodb/selected-tests/blob/master/src/selectedtests/test_mappings/cli.py#L110). That may help the user with a more friendly error response up front as the asynchronous job work planned makes bubbling errors/issues up to users complicated.

It's hard to show that kind of dependency in swagger but additional text in the descriptions for these fields may help.

lydiastepanek: added to the swagger descriptions

PR #14, opened 2019-10-29
TIG-2086: Add build_variant_regex

zamj:


src/selectedtests/task_mappings/mappings.py
```diff
@@ -225,17 +227,19 @@ def _map_tasks_to_files(changed_files: List[str], flipped_tasks: Dict, task_mapp
                 builds_to_task_mappings[cur_task] = cur_flips_for_task + 1
 
 
-def _filter_non_required_distros(builds: List[Build]) -> List[Build]:
+def _filter_non_required_distros(builds: List[Build], build_regex: Pattern) -> List[Build]:
     """
     Filter the distros that aren't required in evergreen.
```
lydiastepanek: I think this method name would be clearer if we replaced "required" with "matching"

src/selectedtests/task_mappings/mappings.py
```diff
@@ -38,6 +38,7 @@ def create_task_mappings(
         file_regex: Pattern,
         module_name: str,
         module_file_regex: Pattern,
+        build_regex: Pattern,
```
lydiastepanek: Could we set the default here to `'!.*'`? That way if someone calls `TaskMappings.create_task_mappings` directly, a default value is set.

It would also be good to test that `TaskMappings.create_task_mappings` succeeds even when `build_regex` is not passed in.

src/selectedtests/task_mappings/mappings.py
zamj: Sure

src/selectedtests/task_mappings/mappings.py
zamj: I've removed the default from click and added the default in to ```create_task_mappings```

src/selectedtests/task_mappings/cli.py
```diff
@@ -107,14 +113,19 @@ def create(
...
 
     mappings = TaskMappings.create_task_mappings(
-        evg_api, evergreen_project, start_date, end_date, file_regex, module_name, module_file_regex
+        evg_api,
+        evergreen_project,
+        start_date,
+        end_date,
+        file_regex,
+        module_name,
+        module_file_regex,
```
dbradf: Should `build_variant_regex` be passed to this call?

tests/selectedtests/conftest.py
```diff
@@ -208,3 +209,8 @@ def _repo(temp_directory):
         return repo
 
     return _repo
+
+
+@pytest.fixture(scope="module")
+def required_builds_regex():
+    return re.compile("!.*")
```
dbradf: It might be a good idea to test with a regex that is different from the parameter default. Otherwise, it could be easy to have a bug where the regex isn't being passed, but using the default.

src/selectedtests/task_mappings/cli.py
zamj: Whoops, forgot to commit those changes. Good catch, fixed

tests/selectedtests/conftest.py
zamj: I've added a test now that makes sure when the regex is different that it is passed down correctly all the way

zamj:
evergreen merge

PR #16, opened 2019-11-01
TIG-1982: Task mappings create api

zamj:


src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -0,0 +1,82 @@
...
+
+                work_item = ProjectTaskMappingWorkItem.new_task_mappings(
+                    project,
+                    work_item_params.get("source_file_regex"),
+                    module,
+                    module_source_file_regex,
+                    work_item_params.get("build_variant_regex"),
+                )
+                if work_item.insert(mongo.task_mappings_queue()):
+                    return jsonify({"custom": f"Work item added for project '{project}'"})
```
lydiastepanek: It would be nice if the 200 status message was the same across test and task mappings, do you mind updating the [test mappings one](https://github.com/mongodb/selected-tests/blob/master/src/selectedtests/app/controllers/project_test_mappings_controller.py#L75) to have the same wording as this one?

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -0,0 +1,82 @@
...
+            evergreen_project = get_evg_project(evg_api, project)
+            if not evergreen_project:
+                abort(404, custom="Evergreen project not found")
+            else:
+                work_item_params = json.loads(request.get_data().decode("utf8"))
+                module = work_item_params.get("module")
+                module_source_file_regex = work_item_params.get("module_source_file_regex")
+                if module is not None and module_source_file_regex is None:
+                    abort(
+                        400,
```
lydiastepanek: It might be helpful to add this 400 response to the Swagger API definition for the endpoint

src/selectedtests/work_items/project_task_mapping_work_item.py
```diff
@@ -0,0 +1,98 @@
...
+        LOGGER.info("Adding new task_mapping work item for project", project=self.project)
+        to_insert = {
+            "created_on": self.created_on,
+            "project": self.project,
+            "source_file_regex": self.source_file_regex,
+        }
+        if self.module:
+            to_insert["module"] = self.module
+            to_insert["module_source_file_regex"] = self.module_source_file_regex
+        if self.build_variant_regex:
```
lydiastepanek: I think it's simpler to just default `module `, `module_source_file_regex `, and `build_variant_regex` to empty strings in the `new_task_mappings` method, rather than only conditionally add them to the work item. When the worker takes work items off the queue, the logic will be simpler if we can assume all fields exist on a work item.

src/selectedtests/app/controllers/project_task_mappings_controller.py
zamj: updated

src/selectedtests/app/controllers/project_task_mappings_controller.py
zamj: added

src/selectedtests/work_items/project_task_mapping_work_item.py
zamj: It matches test-mappings now

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @ns.route("/<project>/task-mappings")
+    @api.param("project", "The evergreen project identifier")
+    class TaskMappingsWorkItem(Resource):
+        @ns.response(200, "Success", response_body)
+        @ns.response(400, "Bad request", response_body)
+        @ns.response(404, "Evergreen project not found", response_body)
+        @ns.response(422, "Work item already exists for project", response_body)
+        @ns.expect(task_mappings_work_item, validate=True)
+        def post(self, project):
```
dbradf: [nit] It could be nice to have the type of `project` and add a description of it to the docstring.

zamj: Updated and added now

zamj:
evergreen merge

PR #20, opened 2019-11-20
TIG-2148: Add helpers for testing workers

dbradf:
This is adding a new helper commands to make it easier to manually test the workers.

### DIFF
--- a/src/selectedtests/datasource/cli.py
+++ b/src/selectedtests/datasource/cli.py
@@ -1,6 +1,5 @@
 """Cli entry point to setup db indexes."""
 import click
-import os
 import structlog
 import logging
 
@@ -17,11 +16,11 @@ def _setup_logging(verbose: bool):
 
 @click.group()
 @click.option("--verbose", is_flag=True, default=False, help="Enable verbose logging.")
+@click.option("--mongo-uri", required=True, type=str, help="Mongo URI to connect to.")
 @click.pass_context
-def cli(ctx, verbose: str):
+def cli(ctx, verbose: bool, mongo_uri: str):
     """Entry point for the cli interface. It sets up the evg api instance and logging."""
     ctx.ensure_object(dict)
-    mongo_uri = os.environ.get("SELECTED_TESTS_MONGO_URI")
     ctx.obj["mongo"] = MongoWrapper.connect(mongo_uri)
 
     _setup_logging(verbose)
@@ -38,4 +37,4 @@ def create_indexes(ctx):
 
 def main():
     """Entry point for setting up selected-tests db indexes."""
-    return cli(obj={})
+    return cli(obj={}, auto_envvar_prefix="SELECTED_TESTS")
--- a/src/selectedtests/helpers.py
+++ b/src/selectedtests/helpers.py
@@ -20,14 +20,13 @@ def get_evg_api():
     return RetryingEvergreenApi.get_api(auth=EvgAuth(evg_user, evg_api_key))
 
 
-def get_mongo_wrapper(mongo_uri: Optional[str] = None) -> MongoWrapper:
+def get_mongo_wrapper() -> MongoWrapper:
     """
     Get an instance of the mongo wrapper based on environment variables.
 
     :return: MongoWrapper instance.
     """
-    if not mongo_uri:
-        mongo_uri = os.environ.get("SELECTED_TESTS_MONGO_URI")
+    mongo_uri = os.environ.get("SELECTED_TESTS_MONGO_URI")
     return MongoWrapper.connect(mongo_uri)
 
 
--- a/src/selectedtests/helpers.py
+++ b/src/selectedtests/helpers.py
@@ -2,7 +2,6 @@
 import os
 import structlog
 import logging
-from typing import Optional
 
 from evergreen.api import RetryingEvergreenApi
 from evergreen.config import EvgAuth
--- a/src/selectedtests/helpers.py
+++ b/src/selectedtests/helpers.py
@@ -2,6 +2,7 @@
 import os
 import structlog
 import logging
+from typing import Optional
 
 from evergreen.api import RetryingEvergreenApi
 from evergreen.config import EvgAuth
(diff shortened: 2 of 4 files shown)


src/selectedtests/datasource/cli.py
```diff
@@ -38,4 +37,4 @@ def create_indexes(ctx):
 
 def main():
     """Entry point for setting up selected-tests db indexes."""
-    return cli(obj={})
+    return cli(obj={}, auto_envvar_prefix="SELECTED_TESTS")
```
dbradf: This is a cool click trick. click can automatically fill in arguments [based on environment variables ](https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables). This means the user can either specify an env var or specify the command line flag.

src/selectedtests/helpers.py
```diff
@@ -19,13 +20,14 @@ def get_evg_api():
     return RetryingEvergreenApi.get_api(auth=EvgAuth(evg_user, evg_api_key))
 
 
-def get_mongo_wrapper() -> MongoWrapper:
+def get_mongo_wrapper(mongo_uri: Optional[str] = None) -> MongoWrapper:
```
dbradf: I switched this to an optional so the calling code can use click's env variables and pass that in, instead of looking up the env var here.

src/selectedtests/work_items/cli.py
```diff
@@ -7,49 +7,116 @@
...
     ctx.ensure_object(dict)
-    ctx.obj["mongo"] = get_mongo_wrapper()
+    ctx.obj["mongo"] = MongoWrapper.connect(mongo_uri)
     ctx.obj["evg_api"] = get_evg_api()
 
     setup_logging(verbose)
 
 
 @cli.command()
+@click.option("--project", type=str, required=True, help="Evergreen project to add.")
```
dbradf: I added this command and a similar command below for task mapping. They both add an item to the appropriate queue to be processed. I find this convenient for local testing. I can add a project to the queue and then run the worker to process that item.

src/selectedtests/work_items/cli.py
```diff
@@ -7,49 +7,116 @@
...
+    evergreen_project = get_evg_project(ctx.obj["evg_api"], project)
+    if not evergreen_project:
+        raise ValueError("Evergreen project not found")
+
+    work_item = ProjectTestMappingWorkItem.new_test_mappings(project, src_regex)
+    work_item.insert(ctx.obj["mongo"].test_mappings_queue())
+
+
+@cli.command()
+@click.option(
```
dbradf: I added this for testing purposes. When actually running against a project, the default of 6 months takes a long time to run. For running locally, I'd rather test 1 week.

src/selectedtests/helpers.py
lydiastepanek: Since this method is no longer called by `src/selectedtests/work_items/cli.py`, is this change still needed? I believe this method is now only called by `src/selectedtests/app/app.py` which does not pass in mongo_uri

src/selectedtests/datasource/cli.py
lydiastepanek: Nice 👍

src/selectedtests/work_items/cli.py
lydiastepanek: Makes sense 👍

src/selectedtests/work_items/cli.py
```diff
@@ -7,49 +7,116 @@
...
+    \f
+    :param ctx: Command Context.
+    :param project: Evergreen project to add to queue.
+    :param src_regex: Regular expression for project source files.
+    """
+    evergreen_project = get_evg_project(ctx.obj["evg_api"], project)
+    if not evergreen_project:
+        raise ValueError("Evergreen project not found")
+
+    work_item = ProjectTestMappingWorkItem.new_test_mappings(project, src_regex)
```
lydiastepanek: Is this erroring? This method should require a `test_file_regex` value be passed in too: https://github.com/mongodb/selected-tests/blob/44fdde29e7110d73692267e6dd2198de9481be8b/src/selectedtests/work_items/project_test_mapping_work_item.py#L55

src/selectedtests/work_items/cli.py
```diff
@@ -7,49 +7,116 @@
...
     setup_logging(verbose)
 
 
 @cli.command()
+@click.option("--project", type=str, required=True, help="Evergreen project to add.")
+@click.option(
+    "--src-regex", type=str, required=True, help="Regular expression for project source files."
+)
+@click.pass_context
+def create_test_mapping(ctx, project: str, src_regex: str):
```
lydiastepanek: Could we rename this to `create_test_mapping_work_item`? At first I thought it did the same thing as the `task-mappings create` command.

src/selectedtests/work_items/cli.py
```diff
@@ -7,49 +7,116 @@
...
 
 
 @cli.command()
+@click.option("--project", type=str, required=True, help="Evergreen project to add.")
+@click.option(
+    "--src-regex", type=str, required=True, help="Regular expression for project source files."
+)
+@click.option("--build-regex", type=str, help="Regular expression for build variants.")
+@click.pass_context
+def create_task_mapping(ctx, project: str, src_regex: str, build_regex: str):
```
lydiastepanek: Could we rename this to `create_task_mapping_work_item` as well?

src/selectedtests/work_items/cli.py
```diff
@@ -7,49 +7,116 @@
...
+
+    work_item = ProjectTaskMappingWorkItem.new_task_mappings(
+        project, src_regex, build_variant_regex=build_regex
+    )
+    work_item.insert(ctx.obj["mongo"].task_mappings_queue())
+
+
+@cli.command()
+@click.option(
+    "--weeks-back", type=int, default=DEFAULT_WEEKS_BACK, help="Number off weeks back to process."
```
lydiastepanek: nit: `Number off` -> `Number of`

src/selectedtests/work_items/cli.py
dbradf: Because these functions get translated by click to be subcommand names, and their parent command is `work_items` adding work_items again is starting to make the command really long. I would prefer to rename the file `work_items_cli.py` to make it clear these are for work_items.

`work-items create-test-mappings` vs `work-items create-test-mapping-work-items`

src/selectedtests/helpers.py
dbradf: yes, fixed.

src/selectedtests/work_items/cli.py
dbradf: fixed.

src/selectedtests/work_items/cli.py
dbradf: fixed.

src/selectedtests/work_items/cli.py
lydiastepanek: Ah, okay. That file rename makes sense. 👍

lydiastepanek: Ah, okay. That file rename makes sense. 👍

dbradf:
evergreen merge

PR #21, opened 2019-11-20
TIG-2148: Add concurrency to task mapping code

dbradf:
This adds concurrency to building the task mappings. Running against 1 weeks worth of data on mongodb-mongo-master, I saw the following timing on my machine (139 versions):

without concurrency: 18.75 minutes
with threads: 2 mins

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 black==19.3b0
 gunicorn==19.9.0
 pydocstyle==4.0.1
+pytest-cov==2.8.1
 pytest-flake8==1.0.4
 pytest==5.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 black==19.3b0
 gunicorn==19.9.0
 pydocstyle==4.0.1
-pytest-cov==2.8.1
 pytest-flake8==1.0.4
 pytest==5.2.0
--- a/src/selectedtests/task_mappings/mappings.py
+++ b/src/selectedtests/task_mappings/mappings.py
@@ -1,6 +1,7 @@
 """Method to create the task mappings for a given evergreen project."""
+from concurrent.futures import ThreadPoolExecutor as Executor
 from datetime import datetime
-from typing import List, Dict, Generator
+from typing import List, Dict, Set
 from tempfile import TemporaryDirectory
 from re import match, compile
 from typing import Pattern
@@ -16,6 +17,7 @@ from selectedtests.evergreen_helper import get_evg_project
 
 LOGGER = get_logger(__name__)
 
+MAX_WORKERS = 16
 SEEN_COUNT_KEY = "seen_count"
 TASK_BUILDS_KEY = "builds"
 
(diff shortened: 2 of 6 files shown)


src/selectedtests/task_mappings/mappings.py
```diff
@@ -62,57 +64,66 @@ def create_task_mappings(
             before_date=before_date,
         )
         log.info("Starting to generate task mappings")
-        project_versions: Generator[Version] = evg_api.versions_by_project(evergreen_project)
+        project_versions = evg_api.versions_by_project_time_window(
```
dbradf: I'm switching this to use the `*_time_window` iterator, which lets us remove some of the logic form this code.

src/selectedtests/task_mappings/mappings.py
```diff
@@ -62,57 +64,66 @@ def create_task_mappings(
...
-                if version.create_time > before_date:
-                    continue
-                if base_repo is None:
+            try:
+                base_repo = _get_evg_project_and_init_repo(evg_api, evergreen_project, temp_dir)
+            except ValueError:
+                LOGGER.warning("Unexpected exception", exc_info=True)
+                raise
+
+            jobs = []
```
dbradf: This is the start of the concurrency code. It is uses futures. The `jobs` variable will keep track of any code we want to execute in the background as futures.

src/selectedtests/task_mappings/mappings.py
```diff
@@ -62,57 +64,66 @@ def create_task_mappings(
...
-
-                if module_name is not None and module_name != "":
-                    cur_module = _get_associated_module(version, module_name)
-                    prev_module = _get_associated_module(prev_version, module_name)
-                    if cur_module is not None and module_repo is None:
-                        module_repo = init_repo(
-                            temp_dir, cur_module.repo, cur_module.branch, cur_module.owner
+                        changed_files = changed_files.union(module_changed_files)
+
+                    jobs.append(
```
dbradf: Here is where we are creating the futures. `exe.submit` adds a jobs to the executors list of jobs to do. The first parameters is the function to execute and the rest of the parameters just get passed to that function when it gets called. It returns a future that we track in the `jobs` list.

src/selectedtests/task_mappings/mappings.py
```diff
@@ -62,57 +64,66 @@ def create_task_mappings(
...
                         )
-
-                    module_changed_files = _get_module_changed_files(
-                        module_repo, cur_module, prev_module, module_file_regex
                     )
-                    changed_files.extend(module_changed_files)
-
-                flipped_tasks = _get_flipped_tasks(prev_version, version, next_version, build_regex)
 
+            for job in jobs:
```
dbradf: Now that all the jobs have been submitted, we wait for them. The  `job.result()` function will wait for the future to complete and return the value the function we gave it returned. So once we get each jobs result, we add it to the map.

src/selectedtests/task_mappings/mappings.py
```diff
@@ -160,47 +171,45 @@ def _get_evg_project_and_init_repo(
     )
 
 
-def _get_filtered_files(diff: DiffIndex, regex: Pattern) -> List[str]:
+def _get_filtered_files(diff: DiffIndex, regex: Pattern) -> Set[str]:
```
dbradf: I changed the files we are looking at to use `Set` instead of `List` since the files should always be unique and we don't care about order.

src/selectedtests/task_mappings/mappings.py
```diff
@@ -265,6 +274,29 @@ def _filter_non_matching_distros(builds: List[Build], build_regex: Pattern) -> L
     return [build for build in builds if match(build_regex, build.display_name)]
 
 
+def _process_evg_version(
```
dbradf: This is pretty much just a wrapper function that is being used for the worker function.

src/selectedtests/work_items/process_work_items.py
```diff
@@ -52,19 +53,32 @@ def process_queued_task_mapping_work_items(
     """
     _clear_in_progress_work(mongo.task_mappings_queue())
     try:
-        while True:
-            out_of_work = _process_one_task_mapping_work_item(
-                evg_api, mongo, after_date, before_date
-            )
-            if out_of_work:
-                break
+        for work_item in _generate_task_mapping_work_items(mongo):
```
dbradf: I started to change this file before I realized it wasn't the bottleneck. It uses a generator to get the work to do, which I think is a little cleaner than looping and then needing to break out of the loop.

src/selectedtests/work_items/process_work_items.py
```diff
@@ -125,6 +131,8 @@ def _run_create_task_mappings(
         mongo.task_mappings().insert_many(transformed_mappings)
     log.info("Finished task mapping work item processing")
 
+    return True
```
dbradf: This function is supposed to return a `bool` and callers trigger off that to know if a work item can be completed.

tests/selectedtests/task_mappings/test_task_mappings.py
```diff
@@ -206,66 +206,6 @@ def test_build_variant_regex_passed_correctly(
 
         assert build_regex == non_matching_filter_mock.call_args[0][1]
 
-    @patch(ns("_get_evg_project_and_init_repo"))
```
dbradf: This was testing the `time_window` logic that is now build into evergreen.py.

src/selectedtests/work_items/process_work_items.py
lydiastepanek: Whoops, nice catch! 

Would you mind adding a `return True` to `_run_create_test_mappings` as well, so the bug is fixed there as well?

src/selectedtests/work_items/process_work_items.py
lydiastepanek: Makes sense, I also think this code is clearer.

My only worry is that there is almost-identical logic in this file as well for test_mappings, will that be refactored when you add concurrency to test mapping code? Or should I create a separate ticket to refactor that logic similarly?

requirements.txt
```diff
@@ -1,5 +1,6 @@
 black==19.3b0
 gunicorn==19.9.0
 pydocstyle==4.0.1
+pytest-cov==2.8.1
```
lydiastepanek: Should we be making this a dev requirement if it's not described in the README and is not included in the evergreen config?

src/selectedtests/work_items/process_work_items.py
dbradf: Updated.

src/selectedtests/work_items/process_work_items.py
dbradf: done

src/selectedtests/task_mappings/mappings.py
```diff
@@ -16,6 +17,7 @@
 
 LOGGER = get_logger(__name__)
 
+MAX_WORKERS = 16
```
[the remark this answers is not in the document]
dbradf: It was random.

I did some testing on my machine, this is what I'm seeing

| MAX_WORKERS | Time (Seconds) |
| --- | --- |
| 16 | 146 |
| 24 | 131 |
| 32 | 108 |
| 64 | 118 |

So around 32 seems to be the sweet spot, I can switch to that.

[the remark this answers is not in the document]
lydiastepanek: Oh, that's a good way to test it. Ok makes sense.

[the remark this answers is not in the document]
lydiastepanek: Oh, that's a good way to test it. Ok makes sense.

dbradf:
evergreen merge

PR #23, opened 2019-11-21
TIG-2053: Add private repo support

dbradf:
This add support for private repos using "Deploy Keys". The user will need to add the selected tests public key as a github deploy key to whatever project that wants to use selected tests. The public/private keys are designed to be specified as env vars to the workers. (I already setup them up in helm).


src/selectedtests/git_helper.py
```diff
@@ -6,7 +6,7 @@
 from git import Commit, Diff, DiffIndex, Repo
 
 
-GITHUB_BASE_URL = "https://github.com"
+GITHUB_BASE_URL = "git@github.com"
```
dbradf: Since we will be using ssh keys, we need to clone via ssh.

cronjobs/process_task_mapping_work_items.sh
```diff
@@ -1,4 +1,10 @@
-#!/bin/sh
+#!/usr/bin/env bash
+
+SSH_DIR=$HOME/.ssh
+SRC_PATH=$(dirname "$0")
```
lydiastepanek: Optional comment: I had to look up what `dirname "$0"` does, I wasn't familiar with this command. Any chance we could rename `SRC_PATH` to `CURRENT_DIR` so it is clearer what this line is doing?

dbradf: I don't like `CURRENT_DIR` because to me that implies cwd (current working directory), which would be the directory the script is being executed from not the directory the script lives in. What about `FILE_DIR`?

lydiastepanek: Ah I see. Yes I like `FILE_DIR`!

dbradf:
evergreen merge

PR #30, opened 2019-12-15
TIG-2001 Daily Model Update

lydiastepanek:
QA'd test mappings with the following steps:
```
# Queue job to create test mappings for mongodb-mongo-master
(selected-tests-3.7.3) ~/m/selected-tests ❯❯❯ curl --verbose -H "Content-Type: application/json" -d '{"source_file_regex":"^src", "test_file_regex":"^jstests","module": "enterprise", "module_source_file_regex":"^src", "module_test_file_regex": "^jstests"}' -X POST http://localhost:8080/projects/mongodb-mongo-master/test-mappings

# Run job to create test mappings for mongodb-mongo-master
(selected-tests-3.7.3) ~/m/selected-tests ❯❯❯ work-items process-test-mappings

# Run daily update script to create test mappings for new commits in mongodb-mongo-master
(selected-tests-3.7.3) ~/m/selected-tests ❯❯❯ test-mappings update
```

QA'd task mappings with the following steps:
```
# Queue job to create task mappings for mongodb-mongo-master
(selected-tests-3.7.3) ~/m/selected-tests ❯❯❯ curl --verbose -H "Content-Type: application/json" -d '{"source_file_regex":"^src", "module": "enterprise", "module_source_file_regex":"^src", "build_variant_regex": "^'!'"}' -X POST http://localhost:8080/projects/mongodb-mongo-master/task-mappings

# Run job to create task mappings for mongodb-mongo-master
(selected-tests-3.7.3) ~/m/selected-tests ❯❯❯ work-items process-task-mappings

# Run daily update script to create task mappings for new versions in mongodb-mongo-master
(selected-tests-3.7.3) ~/m/selected-tests ❯❯❯ task-mappings update
```


src/selectedtests/test_mappings/create_test_mappings.py
```diff
@@ -1,5 +1,6 @@
 """Test Mappings class to create test mappings."""
 import os.path
+import re
```
jimoleary: You had fun with naming conventions for all these classes!

src/selectedtests/test_mappings/create_test_mappings.py
```diff
@@ -27,34 +28,43 @@ def generate_test_mappings(
     evg_api: EvergreenApi,
     evergreen_project: str,
     project_commit_limit: CommitLimit,
-    source_re: Pattern,
-    test_re: Pattern,
+    source_file_regex: str,
```
jimoleary: same comment about regex to pattern

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -23,6 +25,49 @@
...
+    module_name: str = None,
+    module_source_file_regex: str = None,
+    build_variant_regex: str = None,
+) -> Tuple[List[Dict], str]:
+    """
+    Generate task mappings for an evergreen project and its associated module if module is provided.
+
+    :param evg_api: An instance of the evg_api client.
+    :param evergreen_project: The name of the evergreen project to analyze.
+    :param version_limit: The point in time at which to start analyzing versions of the project.
```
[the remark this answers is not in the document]
dbradf: We have had long conversations about this name already. `starting_point` is the exact opposite of how I personally think about this (and the opposite of how the code works, we iterate until we hit this version so calling it starting anything seems confusing). The consensus seemed to be to avoid time based terminology, so we went with `limit`.

src/selectedtests/datasource/mongo_wrapper.py
```diff
@@ -56,3 +56,19 @@ def task_mappings(self):
         :return: task_mappings collection.
         """
         return self.client.selected_tests.task_mappings
+
+    def test_mappings_project_config(self):
```
dbradf: Would it make sense to have just a single project_config collection and keep track of both the test and task states in that collection? It seems like it could be useful to have all the details about a projects in a single location.

src/selectedtests/work_items/process_task_mapping_work_items.py
```diff
@@ -81,23 +93,56 @@ def _run_create_task_mappings(
...
+def update_task_mappings_since_last_commit(evg_api: EvergreenApi, mongo: MongoWrapper):
+    """
+    Update task mappings that are being tracked in the task mappings project config collection.
+
+    :param evg_api: An instance of the evg_api client
+    :param mongo: An instance of MongoWrapper.
+    """
+    LOGGER.info("Updating task mappings")
+    project_cursor = mongo.task_mappings_project_config().find({})
+    for project_config in project_cursor:
```
dbradf: From a design perspective, it might make sense to create a `project_config` domain object that abstracts the project configs. That could remove a lot of the duplication from the `*_work_items` files and keep all the logic for tracking project configs in a single location.

src/selectedtests/task_mappings/create_task_mappings.py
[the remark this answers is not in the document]
jimoleary: The current docstring is: __The point in time at which to start analyzing versions of the project__. So that's very confusing.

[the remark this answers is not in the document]
ancostas: We should make the doc for that argument `How many versions back you want to go.` or such.

src/selectedtests/datasource/mongo_wrapper.py
ancostas: Seconded

ancostas: I'm wondering whether this wrapper is really needed --> you aren't really abstracting away the database, but just hiding exactly which collections are being used, which seems to just make reasoning about things harder for the consumer. Just my 2 cents.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -23,6 +25,49 @@
 TASK_BUILDS_KEY = "builds"
 
 
+def generate_task_mappings(
+    evg_api: EvergreenApi,
```
ancostas: IMO would be simpler if this got refactored at some point to be a `TaskMapper` that is configured with things like `project`, `EvergreenApi`, `Repo` etc and knows how to create `TaskMappings`, and for `TaskMappings` to be a more simple data transfer class that doesn't need to be its own factory --> perhaps only knowing how to serialize itself for storage purposes.

src/selectedtests/work_items/process_task_mapping_work_items.py
```diff
@@ -62,11 +61,24 @@ def _process_one_task_mapping_work_item(
...
     log = LOGGER.bind(project=work_item.project, module=work_item.module)
     log.info("Starting task mapping work item processing for work_item")
-    if _run_create_task_mappings(evg_api, mongo, work_item, after_date, log):
+    if _seed_task_mappings_for_project(evg_api, mongo, work_item, after_date, log):
         work_item.complete(mongo.task_mappings_queue())
 
 
-def _run_create_task_mappings(
+def _create_project_in_task_mappings_config(mongo, work_item, most_recent_version_analyzed):
+    mongo.task_mappings_project_config().insert_one(
```
ancostas: Kind of point in case: we know the database is mongodb at this point, and are using MongoClient's API with some interesting details (i.e. which collection?) being hidden from view.

IMO if you want to abstract mongo away it should look more like something along the lines of `data_service.save(domain_object)` rather than `data_proxy.insert_somewhere(db_representation)`.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -23,6 +25,49 @@
...
+    module_source_file_regex: str = None,
+    build_variant_regex: str = None,
+) -> Tuple[List[Dict], str]:
+    """
+    Generate task mappings for an evergreen project and its associated module if module is provided.
+
+    :param evg_api: An instance of the evg_api client.
+    :param evergreen_project: The name of the evergreen project to analyze.
+    :param version_limit: The point in time at which to start analyzing versions of the project.
+    :param source_file_regex: Pattern to match changed source files against.
```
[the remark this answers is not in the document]
lydiastepanek: True, updated 👍

src/selectedtests/task_mappings/create_task_mappings.py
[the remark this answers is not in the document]
lydiastepanek: I like `The point at which to start analyzing versions of the project`, I will change it to that

src/selectedtests/task_mappings/task_mappings_cli.py
```diff
@@ -97,36 +97,23 @@ def create(
...
 
     LOGGER.info(f"Creating task mappings for {evergreen_project}")
-
-    mappings, _ = TaskMappings.create_task_mappings(
+    mappings, _ = generate_task_mappings(
         evg_api,
         evergreen_project,
         VersionLimit(stop_at_date=after_date),
-        file_regex,
+        source_file_regex,
```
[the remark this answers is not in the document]
lydiastepanek: I agree that changing "regex" to "pattern" makes sense in the functions that explicitly take in a pattern and compile it to a regex, so I updated the `generate_task_mappings` and `generate_task_mappings` functions as you suggested.

Unfortunately, a lot of places we already use the term "regex" to describe a string that will eventually be compiled to a regex. [In the technical design doc](https://docs.google.com/document/d/1azKrJr3babowhr6M8vs9cCTQOmLdUeyqmyVKp_ENSOw/edit#bookmark=id.o1vpxt1c1ik) we stated the params for the CREATE endpoint would be "regex". And [in the Swagger API](https://selected-tests.server-tig.prod.corp.mongodb.com/swagger) we show the params to use the "regex" terminology. So changing it in all the endpoints and CLIs would take a bit of work. If you feel strongly we can make a ticket to do this as a separate piece of work.

src/selectedtests/task_mappings/task_mappings_cli.py
```diff
@@ -97,36 +97,23 @@ def create(
...
         module_name,
-        module_file_regex,
-        build_regex,
+        module_source_file_regex,
+        build_variant_regex,
```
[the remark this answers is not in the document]
lydiastepanek: You're right, the click helpstring was outdated: `build_variant_regex` is an optional argument with no default. What we do not is include all builds in the analysis if no `build_variant_regex` is provided: https://github.com/mongodb/selected-tests/blob/d70e9aac9d97d36e99a2c20bc1853e90d8109324/src/selectedtests/task_mappings/create_task_mappings.py#L278.

src/selectedtests/test_mappings/create_test_mappings.py
lydiastepanek: What do you mean?

src/selectedtests/task_mappings/create_task_mappings.py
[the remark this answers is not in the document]
jimoleary: I assumed the parameter was starting_point based on the current pydoc. 

If '_version_limit_' is the correct name and semantics then the pydoc should be more like 'The point in time to analyzing to.' (or something like that).

src/selectedtests/task_mappings/task_mappings_cli.py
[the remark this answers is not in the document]
jimoleary: If it is already documented (and previously agreed on) then it's fine. Plus any docs produced for the end point will specify the correct type.

src/selectedtests/test_mappings/create_test_mappings.py
jimoleary: Starting non pytest files with the prefix 'test_' in python is a bit of a pain.

src/selectedtests/task_mappings/task_mappings_cli.py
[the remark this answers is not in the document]
jimoleary: :+1:

src/selectedtests/test_mappings/create_test_mappings.py
lydiastepanek: Ah, yes

src/selectedtests/datasource/mongo_wrapper.py
lydiastepanek: I tend to agree with you on this. I was going off of what we do in dag-metrics: https://github.com/10gen/dag-metrics/blob/master/src/dagmetrics/datasource/mongo_wrapper.py. I'm not sure the `MongoWrapper` class adds much, but since we do it in other repos, I think it's something we should discuss as a team?

src/selectedtests/work_items/process_task_mapping_work_items.py
lydiastepanek: I refactored this so that ProjectConfig is a domain object now

src/selectedtests/task_mappings/create_task_mappings.py
lydiastepanek: Perhaps this could work, I do think it would be a big refactor. If you feel strongly we could make a ticket for this.

src/selectedtests/work_items/process_task_mapping_work_items.py
lydiastepanek: Added

src/selectedtests/project_config.py
```diff
@@ -0,0 +1,195 @@
...
+    @classmethod
+    def from_json(cls, json):
+        """
+        Instantiate an instance of TaskConfig from json. Use ProjectConfig.get instead.
+
+        :param json: Json representing project task config in database.
+        :return: An instance of TaskConfig.
+        """
+        return cls(
+            json.get("most_recent_version_analyzed", None),
```
dbradf: [nit] I believe if the given key is not found in the dictionary, `.get` will return `None` if no default is specified. So I don't think you need the 2nd parameter if it is `None`.

src/selectedtests/project_config.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Save a ProjectConfig instance to the db collection.
+
+        :param collection: The collection containing project config documents.
+        """
+        collection.update(
+            {"project": self.project},
+            {
+                "$set": {
+                    "task_config": self.task_config.__dict__,
```
dbradf: Access special methods like `__dict__` of another object can be dangerous. In this case, someone might add internal (private) values to the `task_config` or `test_config` classes that should not be included when exporting the data, but will be if we are using `__dict__`. I personally prefer to be very explicit about what data is being passed to mongo in cases like this.

src/selectedtests/datasource/mongo_wrapper.py
```diff
@@ -1,6 +1,8 @@
 """Classes for accessing mongo collections."""
+from __future__ import annotations
```
dbradf: It looks like the Dockerfile for this projects is using the ubuntu package python3, which I think is python 3.6. This annotation was added in python 3.7. So I think you will need to update the Dockerfile to to use at least python 3.7 for this to work (I recently hit this with the build-baron-tools). The easiest way to do that would be to start basing this image off the python 3.7 debian image.

lydiastepanek: Created a ticket for this: https://jira.mongodb.org/browse/MAKE-1075. For now I removed annotations.

src/selectedtests/project_config.py
```diff
@@ -0,0 +1,208 @@
...
+            )
+        return cls(project, TaskConfig(), TestConfig())
+
+    def save(self, collection):
+        """
+        Save a ProjectConfig instance to the db collection.
+
+        :param collection: The collection containing project config documents.
+        """
+        task_config = {
```
dbradf: [nit] I would pull creating these dictionaries into the `TaskConfig` and `TestConfig` objects. So the code here would just be:
```
project_config = {
  "task_config": task_config.as_dict(),
  "test_config": test_config.as_dict(),
}
```

Maybe even pull that into it's own method to make it symmetric and here is would be:
```
collection.update({"project": self.project}, self.as_dict(), upsert=True)
```

lydiastepanek:
evergreen merge

PR #31, opened 2019-12-19
TIG-2149 Enable default logging in selected-tests project

lydiastepanek:


README.md
```diff
@@ -51,74 +51,13 @@ task-mappings create EVERGREEN_PROJECT_NAME
 ```
 Currently, it can only analyze public git repos. Private repo support is coming in a future version.
 
-Its options are described below.
```
lydiastepanek: Removed this options documentation because having it here is duplicative (it is provided by running `task-mappings create --help`) and easily falls out of date.

src/selectedtests/work_items/process_task_mapping_work_items.py
```diff
@@ -60,10 +61,10 @@ def _process_one_task_mapping_work_item(
     :param mongo: An instance of MongoWrapper.
     :param after_date: The date at which to start analyzing commits of the project.
     """
-    log = LOGGER.bind(project=work_item.project, module=work_item.module)
-    log.info("Starting task mapping work item processing for work_item")
-    if _seed_task_mappings_for_project(evg_api, mongo, work_item, after_date, log):
-        work_item.complete(mongo.task_mappings_queue())
+    with tmp_bind(LOGGER, project=work_item.project, evergreen_module=work_item.module) as log:
```
lydiastepanek: Binding a `module` param raises a `KeyError: "Attempt to overwrite 'module' in LogRecord"` so I renamed it to `evergreen_module`

src/selectedtests/datasource/datasource_cli.py
```diff
@@ -37,7 +32,8 @@ def cli(ctx, verbose: bool, mongo_uri: str):
     ctx.ensure_object(dict)
     ctx.obj["mongo"] = MongoWrapper.connect(mongo_uri)
 
-    _setup_logging(verbose)
+    verbosity = Verbosity.DEBUG if verbose else Verbosity.INFO
```
dbradf: [nit] In other repos, we have been using the click `count` option to set the verbosity (see [here](https://github.com/10gen/build-baron-tools/blob/42c9568632540bd676eef3e0bee78f4af75e7d88/src/bb/bb_cli.py#L143)). Instead of explicitly setting it to a value. The `count` option will set the parameter to the number of times the option was specified. So specifying it once would give INFO, twice DEBUG, etc.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -99,13 +99,7 @@ def create_task_mappings(
...
         :return: An instance of TaskMappings and version_id of the most recent version analyzed.
         """
-        log = LOGGER.bind(
-            project=evergreen_project,
-            module=module_name,
-            version_limit_stop_at_date=version_limit.stop_at_date,
-            version_limit_stop_at_version_id=version_limit.stop_at_version_id,
-        )
-        log.info("Starting to generate task mappings")
+        LOGGER.info("Starting to generate task mappings", version_limit=version_limit.__dict__)
```
dbradf: [nit] I think it would be better to define a `__repr__` method on `version_limit` instead of referencing `__dict__`.

src/selectedtests/datasource/datasource_cli.py
lydiastepanek: I saw that, I thought we could add that later if needed because right I've only found INFO and DEBUG useful. (The workers all use INFO level logging and I use DEBUG when running cli's locally.)

src/selectedtests/task_mappings/create_task_mappings.py
lydiastepanek: Oh whoops, forgot to remove the `__dict__` here

lydiastepanek:
evergreen merge

PR #32, opened 2019-12-19
MAKE-1063 Task mapping creation should increment source_file_seen_cou…

lydiastepanek:
…nt even when no tasks flip


src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -187,23 +186,24 @@ def transform(self) -> List[Dict]:
         task_mappings = []
         for mapping in self.mappings:
```
dbradf: This is unrelated to your change, but could clean up this code some.

All the loops in here look strange to me. There seems to be a pattern of loop over a dictionary, then get the item from the dictionary. Python will do that for us:
```
for source_file, cur_mapping in self.mappings.items():
```

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -187,23 +186,24 @@ def transform(self) -> List[Dict]:
...
-            task_mappings.append(new_mapping)
+            if builds:
+                new_mapping = {
+                    "source_file": mapping,
+                    "project": self.evergreen_project,
+                    "repo": self.repo_name,
+                    "branch": self.branch,
+                    "source_file_seen_count": cur_mappings.get(SEEN_COUNT_KEY),
+                }
+                new_tasks = []
```
dbradf: [nit] I think this could be simplified with list comprehensions. Something like this:
```
new_tasks = [
    {"name": task_name, "variant": build_name, "flip_count": flip_count}
    for task_name, flip_count in build.items()
    for build_name, build in builds
]
```

lydiastepanek: I tried this but I think it's harder to read :(
```
new_tasks = [
    {"name": task_name, "variant": build_name, "flip_count": flip_count}
    for build_name, build in builds.items()
    for task_name, flip_count in build.items()
]
```

lydiastepanek:
evergreen merge

PR #33, opened 2019-12-19
MAKE-1076 Add isort to selected-tests repo

lydiastepanek:


.isort.cfg
```diff
@@ -0,0 +1,8 @@
+[settings]
+force_grid_wrap=0
```
lydiastepanek: Most of these are to enforce compatibility with black: https://github.com/timothycrosley/isort/issues/694#issuecomment-564261886

evergreen.yml
```diff
@@ -60,3 +60,14 @@ tasks:
           . venv/bin/activate
 
           pydocstyle src
+
+- name: check_isort
```
dbradf: We might want to consider using the [pytest plugin ](https://pypi.org/project/pytest-isort/)

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -1,20 +1,19 @@
...
 from evergreen.manifest import ManifestModule
-from git import Repo, DiffIndex
-from structlog import get_logger
-from tempfile import TemporaryDirectory
-
+from git import DiffIndex, Repo
 from selectedtests.evergreen_helper import get_evg_project
 from selectedtests.git_helper import get_changed_files, init_repo
 from selectedtests.task_mappings.version_limit import VersionLimit
+from structlog import get_logger
```
dbradf: It looks like `isort` is treating the local library as a 3rd party library. Do you know if there is a way to tell it the difference?

lydiastepanek: Adding the `known_first_party` config option seems to have solved this

lydiastepanek:
evergreen merge

PR #34, opened 2020-01-02
MAKE-1070 Task mapping creation should set repo value to module's repo name for module-related task mappings

lydiastepanek:


src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -215,15 +216,20 @@ def _get_evg_project_and_init_repo(
     )
 
 
-def _get_filtered_files(diff: DiffIndex, regex: Pattern) -> Set[str]:
+def _get_filtered_files(diff: DiffIndex, regex: Pattern, repo_name: str) -> Set[str]:
```
zamj: This return type has changed

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @ancostas @dbradf @lydiastepanek @jimoleary @varsha804 @zamj
+* @ancostas @dbradf @lydiastepanek @jimoleary @varsha804 @zamj
```
dbradf: I don't see any changes to this file?

tests/selectedtests/task_mappings/test_create_task_mappings.py
```diff
@@ -86,11 +90,14 @@ def test_module_source_files_included(
         only_version_analyzed = evg_api_mock.versions_by_project.return_value[1]
 
         associated_module_mock.return_value = None
-        module_changed_mock.return_value = {"module_file"}
-        filtered_mock.return_value = {f"file{i}" for i in range(2)}
+        module_changed_mock.return_value = {ChangedFile("module_file", "my_module_repo")}
+        filtered_files_mock.return_value = {
```
dbradf: Why change all these assignments from using comprehensions to creating each instance manually? The comprehensions tend to be more flexible (e.g. if we wanted to create 5 values instead of 2 it is a 1 character change). Also if we change how the value is represented (like it this case), we only have to change how it is created in one place instead of changing each instance.

.github/CODEOWNERS
lydiastepanek: Reverted. (I wasn't sure why @varsha804 wasn't getting added as a Reviewer, so I tried removing the new line at the end of this file. I think actually the problem was that I needed to add her as a [collaborator](https://github.com/mongodb/selected-tests/settings/collaboration).)

tests/selectedtests/task_mappings/test_create_task_mappings.py
lydiastepanek: I'm happy to change it back, I just find it easier to read hard coded variables rather than ones created by loops. So if there were fewer than three items being created, I removed the loop.

dbradf: We are doing it enough in this file that it might make sense to just have a helper function that creates some number of `ChangedFile`s. Then just have all the instances you changed just call the function. That way if we make changes in the future, we just have 1 place to change.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -21,6 +22,7 @@
 MAX_WORKERS = 32
 SEEN_COUNT_KEY = "seen_count"
 TASK_BUILDS_KEY = "builds"
+ChangedFile = namedtuple("ChangedFile", ["file_name", "repo_name"])
```
ancostas: For your consideration:
```
class ChangedFile(NamedTuple):
    file_name: str
    repo_name: str
```

lydiastepanek: I think a named tuple takes less code. If we add more attributes later, a class would make it easier to manage them, but it's easy to change to a class later.

lydiastepanek:
evergreen merge

PR #38, opened 2020-01-16
MAKE-1075, MAKE-1081: Integrate mypy and move to python 3.7

ancostas:


src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -47,8 +48,12 @@ def generate_task_mappings(
     :return: An instance of TestMappingsResult and the most recent version analyzed during analysis.
     """
     source_re = re.compile(source_file_pattern)
-    module_source_re = None
+    module_source_re = re.compile("")
```
ancostas: I think match gets called on this downstream no matter what --> or I've missed something.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -172,6 +177,8 @@ def create_task_mappings(
                 changed_files, flipped_tasks = job.result()
                 _map_tasks_to_files(changed_files, flipped_tasks, task_mappings)
 
+        if branch is None:
+            raise RuntimeError("Unable to identify branch")
```
ancostas: I think this guy is always needed, or?

src/selectedtests/test_mappings/create_test_mappings.py
```diff
@@ -59,19 +61,19 @@ def generate_test_mappings(
...
         )
 
-        if module_name:
+        if module_name is not None:
+            if module_source_file_pattern is None or module_test_file_pattern is None:
+                raise ValueError(
+                    "You must specify module_source_file_pattern and module_test_file_pattern when passing a module"
+                )
+            module_source_re = re.compile(module_source_file_pattern)
+            module_test_re = re.compile(module_test_file_pattern)
```
ancostas: I think the logic is basically equivalent, and makes mypy happy.

tests/selectedtests/app/conftest.py
```diff
@@ -1,16 +0,0 @@
...
-
-@pytest.fixture()
-def app_client():
-    """
-    Client for the flask web with mocked endpoints.
-    """
-    mongo = MagicMock()
-    evg_api = MagicMock()
-    client = app.create_app(mongo, evg_api).test_client()
-    return client
```
ancostas: Mypy didn't like the duplicate module name, so I moved this into the other conftest.

ancostas:
I did some minor refactoring based on mypy complaining --> I tried to call out wherever code changes were necessary and perhaps error prone. Let me know what you think!

src/selectedtests/test_mappings/create_test_mappings.py
lydiastepanek: You can remove this line as well, right? `most_recent_module_commit = None`

src/selectedtests/test_mappings/create_test_mappings.py
```diff
@@ -33,7 +35,7 @@ def generate_test_mappings(
     source_file_pattern: str,
     test_file_pattern: str,
     module_name: str = None,
-    module_commit_limit: CommitLimit = None,
+    module_commit_limit: CommitLimit = CommitLimit(),
```
lydiastepanek: I think this should default to None

src/selectedtests/datasource/mongo_wrapper.py
```diff
@@ -1,4 +1,5 @@
 """Classes for accessing mongo collections."""
+from __future__ import annotations
```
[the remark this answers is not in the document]
ancostas: Upgraded to `python:3.7-slim-buster`

src/selectedtests/test_mappings/create_test_mappings.py
ancostas: A test fails when I remove `most_recent_module_commit = None`.

Dockerfile
```diff
@@ -1,13 +1,4 @@
...
-RUN set -xe \
-    && apt-get update \
-    && apt-get install -y curl gcc g++ \
-    && apt-get install -y python3-dev \
-    && apt-get install -y libffi-dev \
-    && apt-get install -y libssl-dev \
-    && apt-get install -y python3-pip \
-    && apt-get install -y git
-RUN pip3 install --upgrade pip
+FROM python:3.7-slim-buster
```
dbradf: This will probably break the deployed version since git (and maybe curl) are not longer installed. This project requires git being installed.

dbradf: Also, MAKE-1075 is open to do this.

requirements.txt
```diff
@@ -4,3 +4,4 @@ pydocstyle==4.0.1
 pytest-flake8==1.0.4
 pytest-isort==0.3.1
 pytest==5.2.0
+pytest-mypy~=0.4.2
```
dbradf: Tools like this we might want to hard code to a specific version so that changes between version don't cause new errors.

src/selectedtests/app/app.py
```diff
@@ -42,7 +42,7 @@ def create_app(mongo: MongoWrapper, evg_api: EvergreenApi) -> Flask:
     return app
 
 
-def main():
+def main() -> Flask:
```
dbradf: I believe this ticket was originally prioritized below MAKE-1085 (switch to fastAPI), since a chunk of typing will probably change with that ticket.

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -65,7 +67,7 @@ class TaskMappings(Resource):
         @ns.response(400, "Bad request")
         @ns.response(404, "Evergreen project not found")
         @ns.expect(parser)
-        def get(self, project: str):
+        def get(self, project: str) -> Response:  # type: ignore
```
dbradf: Why all the `type: ignore`s?

Dockerfile
ancostas: Good catch, I've added git/curl back in. Is curl just there as a transitive dependency for git though, or do we use it somewhere else?

@lydiastepanek I think I missed MAKE-1075 since it was in this sprint assigned to you and there's no link in the tix. Do you want to do a quick PR for that ticket and I'll rebase, or should I just take that one as well and do both in this PR?

src/selectedtests/test_mappings/create_test_mappings.py
lydiastepanek: Talked IRL, nvm

Dockerfile
lydiastepanek: Feel free to take MAKE-1075 and assign to yourself.

requirements.txt
ancostas: I've made it a fixed version, although I'm wondering whether semver should protect us from errors when automatically pulling patches.

src/selectedtests/app/app.py
ancostas: My bad, I took the most straightforward looking ticket I saw in order to take a break from cedar stuff. The Web API layer is pretty slim so hopefully wont be too bad.

src/selectedtests/app/controllers/project_task_mappings_controller.py
ancostas: mypy doesn't like that we don't return anything (when calling `abort`).

Dockerfile
dbradf: git is definitely required. I'm not sure about curl. I know we need it in other tools (to install node, I think), but we might not need it here.

ancostas:
evergreen merge

PR #39, opened 2020-01-22
MAKE-1085: Switch to FastAPI framework

ancostas:
https://selected-tests.server-tig.staging.corp.mongodb.com/swagger


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-slim-buster
+FROM python:3.7-buster
```
ancostas: slim doesn't have some stuff needed for building `httptools`'s wheels.

scripts/server.sh
```diff
@@ -3,4 +3,4 @@
 export LC_ALL=C.UTF-8
 export LANG=C.UTF-8
 
-gunicorn --bind 0.0.0.0:8080 --workers 3 wsgi
+uvicorn --host 0.0.0.0 --port 8080 --workers 1 selectedtests.app.app:app
```
ancostas: Switching to one worker since this is best practice per Kanopy docs.

ancostas:
There is some more work needed here to fix-up the tests. Right now they expect the app to pass around its dependencies all over the place.

ancostas:
Ready now.

src/selectedtests/app/app.py
```diff
@@ -1,53 +1,43 @@
...
+        project_task_mappings_controller.router,
+        prefix="/projects/{project}/task-mappings",
+        tags=["projects"],
+    )
+    app.include_router(
+        project_test_mappings_controller.router,
+        prefix="/projects/{project}/test-mappings",
+        tags=["projects"],
+    )
+    app.state.db = mongo_wrapper or get_mongo_wrapper()
```
lydiastepanek: I would prefer `app.state.db = mongo_wrapper` where mongo_wrapper is instantiated outside of the `create_app` function and passed in as a required argument.

src/selectedtests/app/dependencies.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def get_db(request: Request) -> MongoWrapper:
+    """
+    Get the configured database for the application.
+
+    :param request: The request needing access to the database.
+    :return: The database.
+    """
+    return request.app.state.db
```
lydiastepanek: Would be curious to know how @dbradf feels about using app state to store the `db` variable. I think I am ok with this.

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -1,130 +1,106 @@
...
-        required=True,
-    )
-    parser.add_argument(
-        "threshold",
-        type=Decimal,
-        location="args",
-        help="Minimum threshold desired for flip_count / source_file_seen_count ratio",
+
+@router.post(path="", response_model=CustomResponse)
+def post(
```
lydiastepanek: The 404 and 422 responses for the post methods are not showing up in swagger

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -1,130 +1,106 @@
...
+    """Model for task mapping responses."""
 
-def add_project_task_mappings_endpoints(
-    api: Api, mongo: MongoWrapper, evg_api: EvergreenApi
-) -> None:
+    task_mappings: List = []
+
+
+@router.get(path="", response_model=TaskMappingsResponse)
+def get(
```
lydiastepanek: The 404 response for the get methods is not showing up in swagger

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -1,130 +1,106 @@
...
+    :param db: The database.
+    :param work_item_params: The work items to enqueue.
+    :param project: The evergreen project identifier.
+    """
+    module = work_item_params.module
+    module_source_file_regex = work_item_params.module_source_file_regex
+    if module and not module_source_file_regex:
+        raise HTTPException(
+            status_code=400,
+            detail="The module_source_file_regex param is required if "
```
lydiastepanek: Out of curiosity, why use `detail` for errors rather than `custom`?

tests/selectedtests/app/controllers/test_project_task_mappings_controller.py
```diff
@@ -85,108 +87,67 @@ def test_POST_work_item_inserted(
...
-    )
+    response = app_client.post(f"/projects/{project}/task-mappings", json=test_params)
     assert response.status_code == 200
-    assert response.get_json()["custom"] == f"Work item added for project '{project}'"
+    assert response.json()["custom"] == f"Work item added for project '{project}'"
 
 
 @patch(ns("ProjectTaskMappingWorkItem"))
-@patch(ns("get_evg_project"))
-def test_POST_work_item_inserted_with_incorrect_params(
```
lydiastepanek: Why remove this test?

src/selectedtests/app/app.py
```diff
@@ -1,53 +1,43 @@
...
-)
-from selectedtests.app.swagger_api import Swagger_Api
 from selectedtests.datasource.mongo_wrapper import MongoWrapper
 from selectedtests.helpers import get_evg_api, get_mongo_wrapper
 
-DEFAULT_PORT = 8080
-
 
-def create_app(mongo: MongoWrapper, evg_api: EvergreenApi) -> Flask:
+def create_app(mongo_wrapper: MongoWrapper = None, evg_api: EvergreenApi = None) -> FastAPI:
```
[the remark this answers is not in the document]
ancostas: Removed the defaults and `None` override.

src/selectedtests/app/controllers/project_task_mappings_controller.py
ancostas: Good catch.

src/selectedtests/app/controllers/project_task_mappings_controller.py
ancostas: That's the interface for fast api's `HTTPException` it accepts `statuscode`, `detail` and `headers`.

tests/selectedtests/app/controllers/test_project_task_mappings_controller.py
ancostas: FastAPI coerces `3` to become `"3"` since the schema says it needs to be a string. So this won't succeed to fail.

src/selectedtests/app/dependencies.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Get the configured Evergreen API client for the application.
+
+    :param request: The request needing the Evergreen API client.
+    :return: The Evergreen API client.
+    """
+    return request.app.state.evg_api
+
+
+def retrieve_evergreen_project(project: str, api: EvergreenApi = Depends(get_evg)) -> Project:
```
dbradf: This also seem strange to have a DI. This is a function, I'm not sure why it would be included as part of the DI.

src/selectedtests/app/dependencies.py
dbradf: This is probably ok.

tests/selectedtests/app/controllers/test_project_task_mappings_controller.py
lydiastepanek: Well could you keep the test and change the failure to whatever the FastAPI failure response is?

tests/selectedtests/app/test_app.py
```diff
@@ -1,17 +1,17 @@
-from flask import testing
+from starlette.testclient import TestClient
 
 
-def test_swagger_endpoint(app_client: testing.FlaskClient):
+def test_swagger_endpoint(app_client: TestClient):
     """
-    Test /health endpoint
+    Test swagger endpoint
```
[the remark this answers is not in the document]
ancostas: Probably not.

tests/selectedtests/app/controllers/test_project_task_mappings_controller.py
ancostas: FastAPI doesn't fail, it just coerces 3 to "3" --> which is a valid string and can be used as a regex pattern.

src/selectedtests/app/main.py
```diff
@@ -0,0 +1,12 @@
+"""Application to serve API of selected-tests service."""
```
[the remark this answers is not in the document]
ancostas: Good point, will remove.

src/selectedtests/app/dependencies.py
```diff
@@ -0,0 +1,53 @@
...
+
+from evergreen import EvergreenApi, Project
+from fastapi import Depends, HTTPException
+from starlette.requests import Request
+
+from selectedtests.datasource.mongo_wrapper import MongoWrapper
+from selectedtests.evergreen_helper import get_evg_project
+
+
+def changed_files_parser(changed_files: str) -> List[str]:
```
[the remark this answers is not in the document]
ancostas: Common query parameter parsing/handling is one of the examples given by FastAPI as to when to use their dependency system.

tests/selectedtests/app/controllers/test_project_task_mappings_controller.py
lydiastepanek: Oh I see now, makes sense

src/selectedtests/app/dependencies.py
ancostas: Same for this one -- every controller method that needs `project` is doing the same business logic --> seemed like they just depend on the project existing and then want it passed to them. Will follow up offline to discuss further.

src/selectedtests/app/helpers.py
```diff
@@ -0,0 +1,31 @@
+"""Helper functions for selected tests API."""
```
dbradf: [nit] I'm not a fan of files named "helpers". It would be nice if this was a more descriptive name or these functions were in more descriptive places.

tests/selectedtests/app/test_app.py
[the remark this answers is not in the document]
dbradf: Can we delete them?

asgi.py
```diff
@@ -0,0 +1,4 @@
+from selectedtests.app.app import create_app
+from selectedtests.helpers import get_evg_api, get_mongo_wrapper
+
+app = create_app(get_mongo_wrapper(), get_evg_api())
```
dbradf: [nit] should this file be in the root of the project? 

We could put it in `src/selectedtests/app/app.py` basically replacing the `main` function there.

ancostas:
evergreen merge

PR #40, opened 2020-02-10
Make scripts executable

lydiastepanek:


ancostas:
Its pretty obvious what happened here based on the commit message, but I was wondering: is there actually a way to review file permissions in github, or do you just need to clone and review locally?

lydiastepanek:
evergreen merge

lydiastepanek:
@ancostas pretty sure that `100644 → 100755` in github means the file mode has changed from 644 to 755: https://stackoverflow.com/questions/16748160/what-does-it-mean-when-git-diff-shows-mode-changes/16748188

PR #42, opened 2020-02-18
MAKE-1106 Task mappings should only process display tasks for tasks split into display and execution tasks.

jimoleary:


jimoleary:
This is my first naive / simple cut at this. I still need to look more closely to ensure that this is correct wrt to results.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -419,7 +420,14 @@ def _create_task_map(tasks: List[Task]) -> Dict:
     :param tasks: List of tasks to map.
     :return: Dictionary of tasks by display_name.
     """
-    return {task.display_name: task for task in tasks}
+    execution_tasks_map = {
```
dbradf: Does this need to be a map? Would it make more sense for it to be a set?

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -419,7 +420,14 @@ def _create_task_map(tasks: List[Task]) -> Dict:
...
+        task: task
+        for task in itertools.chain.from_iterable(
+            task.json["execution_tasks"] for task in tasks if "execution_tasks" in task.json
```
dbradf: [nit] There might be some logic here we would want to pull into evergreen.py at some point. I can imagine other places where this might be useful.

src/selectedtests/task_mappings/create_task_mappings.py
jimoleary: For the larger builds (containing large numbers of large display tasks) a map is 14 times faster for more or less the same syntax.

src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -419,7 +420,14 @@ def _create_task_map(tasks: List[Task]) -> Dict:
     :param tasks: List of tasks to map.
     :return: Dictionary of tasks by display_name.
     """
-    return {task.display_name: task for task in tasks}
+    execution_tasks_map = {
+        task: task
+        for task in itertools.chain.from_iterable(
```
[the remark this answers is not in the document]
jimoleary: Fair point, I've changed the outer variable name to execution_task_id which describes it exactly.

src/selectedtests/task_mappings/create_task_mappings.py
jimoleary: I'm going to leave that to another PR. evergreen.py doesn't expose execution_tasks as a top level attribute at the moment.

jimoleary:
> Any idea what is going on with evergreen?

evergreen occassionally fails, restarting or pushing again ususally resolves transient issues.

src/selectedtests/task_mappings/create_task_mappings.py
dbradf: A map is faster than a set for this? That seems really surprising.

jimoleary: I went back and looked at my testing and it looks like I tested lists twice rather than list, sets and dicts.

A set is more appropriate / clear for this (and the same speed) so I going to change to that.

PR #44, opened 2020-02-21
MAKE-1120 specify resources.

jimoleary:
The cron jobs are set as __Burstable__ as they will run again later. If the requests stanza is removed then they would be __Guaranteed__ and would have higher priority.

The prod configuration is also __Burstable__ , with the same (high) limits as the cronjobs, as we sometimes update the database in this container (although maybe we shouldn't be doing that).

### DIFF
--- a/cronjobs.yml
+++ b/cronjobs.yml
@@ -11,38 +11,38 @@ cronJobs:
     command: ["/bin/sh", "cronjobs/process_test_mapping_work_items.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 10m
-        memory: 5Mi
+        cpu: "10m"
+        memory: "5Mi"
       limits: # maximum allowed resources
-        cpu: 20m
-        memory: 10Mi
+        cpu: "1.5"
+        memory: "600Mi"
   - name: process-task-mappings
     schedule: "0 2 * * *"
     command: ["/bin/sh", "cronjobs/process_task_mapping_work_items.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 10m
-        memory: 5Mi
+        cpu: "10m"
+        memory: "5Mi"
       limits: # maximum allowed resources
-        cpu: 20m
-        memory: 10Mi
+        cpu: "2.5"
+        memory: "1.5Gi"
   - name: update-test-mappings
     schedule: "0 3 * * *"
     command: ["/bin/sh", "cronjobs/update_test_mappings.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 400m
-        memory: 800Mi
+        cpu: "900m"
+        memory: "500Mi"
       limits: # maximum allowed resources
-        cpu: 600m
-        memory: 1.2Gi
+        cpu: "1"
+        memory: "750Mi"
   - name: update-task-mappings
     schedule: "0 4 * * *"
     command: ["/bin/sh", "cronjobs/update_task_mappings.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 1
-        memory: 2.25Gi
+        cpu: "1.1"
+        memory: "800Mi"
       limits: # maximum allowed resources
-        cpu: 1.5
-        memory: 3.375Gi
+        cpu: "1.2"
+        memory: "1Gi"
--- a/cronjobs.yml
+++ b/cronjobs.yml
@@ -11,38 +11,38 @@ cronJobs:
     command: ["/bin/sh", "cronjobs/process_test_mapping_work_items.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 500m
-        memory: 6Gi
+        cpu: 10m
+        memory: 5Mi
       limits: # maximum allowed resources
-        cpu: 2
-        memory: 12Gi
+        cpu: 20m
+        memory: 10Mi
   - name: process-task-mappings
     schedule: "0 2 * * *"
     command: ["/bin/sh", "cronjobs/process_task_mapping_work_items.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 500m
-        memory: 6Gi
+        cpu: 10m
+        memory: 5Mi
       limits: # maximum allowed resources
-        cpu: 2
-        memory: 12Gi
+        cpu: 20m
+        memory: 10Mi
   - name: update-test-mappings
     schedule: "0 3 * * *"
     command: ["/bin/sh", "cronjobs/update_test_mappings.sh"]
     resources: # Burstable
       requests: # guaranteed amount of resources
-        cpu: 500m
-        memory: 6Gi
+        cpu: 400m
+        memory: 800Mi
       limits: # maximum allowed resources
-        cpu: 2
-        memory: 12Gi
+        cpu: 600m
+        memory: 1.2Gi
   - name: update-task-mappings
     schedule: "0 4 * * *"
     command: ["/bin/sh", "cronjobs/update_task_mappings.sh"]
     resources: # Burstable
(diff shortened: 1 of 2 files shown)


cronjobs.yml
```diff
@@ -9,12 +9,40 @@ cronJobs:
   - name: process-test-mappings
     schedule: "0 1 * * *"
     command: ["/bin/sh", "cronjobs/process_test_mapping_work_items.sh"]
+    resources: # Burstable
+      requests: # guaranteed amount of resources
+        cpu: 500m
+        memory: 6Gi
```
dbradf: These requests and limits are huge. Why do we need so much memory?

environments/prod.yml
```diff
@@ -13,3 +13,10 @@ prometheusRules:
     annotations:
       description: "High number of bad responses coming from selected-tests service"
       summary: "Selected tests service giving high number of bad responses over the last hour (current number: {{ $value }}). More information can be seen here: https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4"
+resources: # Burstable
+  requests: # guaranteed amount of resources
+    cpu: 500m
+    memory: 6Gi
+  limits: # maximum allowed resources
+    cpu: 2
+    memory: 12Gi
```
ancostas: This seems like a lot of CPU and memory. Looking over the [last 7 days](https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4&from=now-7d&to=now) CPU has been below 0.007 and memory below 0.1 Gi. Should we perhaps do something closer to that in order to make these pods easier to schedule?

cronjobs.yml
```diff
@@ -9,12 +9,40 @@ cronJobs:
...
   - name: update-task-mappings
     schedule: "0 4 * * *"
     command: ["/bin/sh", "cronjobs/update_task_mappings.sh"]
+    resources: # Burstable
+      requests: # guaranteed amount of resources
+        cpu: 500m
+        memory: 6Gi
+      limits: # maximum allowed resources
+        cpu: 2
+        memory: 12Gi
```
ancostas: Same here, the update cronjobs seem to be staying below 1 CPU and 3 Gi. The process jobs aren't even showing up due to being very short-lived.

environments/prod.yml
jimoleary: At the moment, we have been running the reseed the database update in this container. So it would need to be able to burst up to what ever the limits are for that scenario.

The normal web service should be use very little resources.

We should probably not seed the database from this container.

cronjobs.yml
jimoleary: I misread the maxes in TIG-2051 as the recommendations, I'll recalibrate.

cronjobs.yml
jimoleary: The normal case is when they run an update. That is a delta so they tend to be small. 

If it needs to be run for the last 6 months then the requirements are much greater. I can lower them and we can see if we hit the limits.

jimoleary:
The dashboard is [here](https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4&from=now-7d&to=now&refresh=5s).

cronjobs.yml
```diff
@@ -9,12 +9,40 @@ cronJobs:
...
+        memory: "5Mi"
+      limits: # maximum allowed resources
+        cpu: "2.5"
+        memory: "1.5Gi"
   - name: update-test-mappings
     schedule: "0 3 * * *"
     command: ["/bin/sh", "cronjobs/update_test_mappings.sh"]
+    resources: # Burstable
+      requests: # guaranteed amount of resources
+        cpu: "900m"
```
lydiastepanek: Why do the `update` cronjobs need more resources than the `process` cronjobs? Since they only run on 1 day of data, as opposed to 6 months, shouldn't the `updates` use fewer resources?

cronjobs.yml
```diff
@@ -9,12 +9,40 @@ cronJobs:
   - name: process-test-mappings
     schedule: "0 1 * * *"
     command: ["/bin/sh", "cronjobs/process_test_mapping_work_items.sh"]
+    resources: # Burstable
+      requests: # guaranteed amount of resources
+        cpu: "10m"
```
lydiastepanek: Out of curiosity, why add quotes to the metrics?

environments/prod.yml
```diff
@@ -13,3 +13,10 @@ prometheusRules:
     annotations:
       description: "High number of bad responses coming from selected-tests service"
       summary: "Selected tests service giving high number of bad responses over the last hour (current number: {{ $value }}). More information can be seen here: https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4"
+resources: # Burstable
+  requests: # guaranteed amount of resources
+    cpu: "5m"
+    memory: "30Mi"
```
lydiastepanek: I know very little about what are standard cpu/memory requests, but 5m cpu and 30Mi memory look low to me. (I say this only because the [kanopy docs](https://github.com/10gen/kanopy-docs/blob/master/docs/production_checklist.md#resource-requestslimits) put 100m cpu and 192Mi memory in their example.) Might we set these to 100m cpu and 192Mi memory, just to be safe?

cronjobs.yml
jimoleary: This is explained in the [application resources docs here](https://github.com/10gen/kanopy-docs/blob/f4cbf669a54574420248792ad3c78b9e4fa96f81/docs/application_resources.md).

> Requests are the values that you expect each container to use and is used by the scheduler to place the container (Pod) in the appropiate node given the available resources on it. Limits are a maximum cap of resources that a container can consume. If utilization exceeds the requested values it will cause the container's CPU use to be throttled, or to be evicted with an Out Of Memory (OOM) error. More information about requests and limits can be found in the official docs here.

The update jobs typically use the same amount of resources every time so that is what the requests value is set to ( have a look at [grafana](https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4&refresh=5s&from=now-7d&to=now)). The limit is set to some theoretical max.

The process jobs typically use very little resources (as they don't do much and are usually so low or so fast that we can't even see some of the resources [grafana](https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4&refresh=5s&from=now-7d&to=now)) so the requests values are set low. However, when they are doing a full reload they then use a lot of resources so the limit is set quite high.

This seems counter-intuitive to what you would expect, but I think it makes sense.

environments/prod.yml
jimoleary: From [Grafana](https://grafana.corp.mongodb.com/d/zenZa6oWk/selected-tests-service?orgId=4), 100m cpu and 192Mi would be over provisioning by about 50x (at the moment). The request resources used by the web app are incredibly low right now.

I would like to provision the application low and increase it later if there is an issue.

I don't think there is any harm in setting these values lower than the suggested values (as they are also multiples of the actual required resources). It may also help us validate that the application limits are being enforced.

cronjobs.yml
lydiastepanek: Makes sense 👍

cronjobs.yml
lydiastepanek: Ah, I see, that makes sense. Thanks for explaining.

environments/prod.yml
lydiastepanek: But we don't have any alerts set up on application limits right now, so we won't know right away if there is an issue. I'd prefer to overprovision up front since it is a web app and traffic will increase once this project is live, but I do not feel super strongly about this, so am fine with it as-is.

cronjobs.yml
jimoleary: This comment has now been addressed. I'm going to merge (and thus deploy the changes). We can revisit the limits if there is an issue.

PR #45, opened 2020-03-04
TIG-2430 Graph distribution of thresholds for task_mappings and test_mappings

lydiastepanek:
@jimoleary and I paired to create Jupyter notebooks that would display the distribution of thresholds.

Distribution of task mappings thresholds:
Screen Shot 2020-03-04 at 12 18 40 PM

Distribution of test mappings thresholds:
Screen Shot 2020-03-04 at 12 19 13 PM


notebooks/.ipynb_checkpoints/task_mapping_threshold-checkpoint.ipynb
```diff
@@ -0,0 +1,96 @@
+{
```
jimoleary: can we add _notebooks/.ipynb_checkpoints_ to the .gitignore and remove these files?

requirements.txt
```diff
@@ -4,4 +4,7 @@ pydocstyle==4.0.1
 pytest-flake8==1.0.4
 pytest-isort==0.3.1
 pytest==5.2.0
-pytest-mypy==0.4.2
\ No newline at end of file
+pytest-mypy==0.4.2
```
jimoleary: These should be requirements-dev.txt (assuming we want to follow that pattern) as they will likely make any container (much) larger!

README.md
```diff
@@ -185,7 +185,17 @@ are joined to _test_mappings_ documents  _test_files_ field through the source_f
 	"test_mpapping_id" : ...,
 	"test_file_seen_count" : 2
 }
+
+### Analyze distrubution of thresholds
+The distribution of the thresholds for test_mappings and task_mappings can be
+seen by running the jupyter notebooks in the `notebooks` directory. To run
+them, set the SELECTED_TESTS_MONGO_URI environment variable to the database you
+would like to analyze and run jupyter notebooks:
+```
```
jimoleary: maybe  add `pip install -r requirements-dev.txt ` (see my later comment).

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @evergreen-ci/dev-prod-dag
+* @10gen/dev-prod-dag
```
dbradf: You probably want `@mongodb/dev-prod-dag`. Unfortunately, the way github works, we have a separate team for each org (but it is 'dev-prod-dag' in all the orgs).

requirements.txt
dbradf: If at some point we want to switch to poetry, we can replace this with just the .toml file.

lydiastepanek:
@dbradf This is against 6 months of data but we updated the AC of [TIG-2413](https://jira.mongodb.org/browse/TIG-2413) to re-run the distribution once that ticket is done and we are using 3 years of data.

requirements.txt
lydiastepanek: @jimoleary as discussed IRL, setup.py contains the production requirements, and requirements.txt contains the dev dependencies, so the containers will not install the requirements.txt requirements.

README.md
lydiastepanek: I would argue the setup steps should only be described once in the README, which happens here: https://github.com/mongodb/selected-tests#set-up-environment. Do you think this is clear enough, or should be rearranged?

jimoleary: no, this is fine.

requirements.txt
jimoleary: agreed, this relates to my misunderstanding of the structure. Please ignore the original comment.

PR #46, opened 2020-03-05
TIG-2413 Evaluate creating test and task mappings using 3 years' worth of data rather than 6 months

lydiastepanek:
Planning on re-seeding the database using this logic.

### DIFF
--- a/src/selectedtests/task_mappings/create_task_mappings.py
+++ b/src/selectedtests/task_mappings/create_task_mappings.py
@@ -138,7 +138,7 @@ class TaskMappings:
                         branch = version.branch
                         repo_name = version.repo
 
-                    LOGGER.info("Processing mappings for version", version=version.version_id)
+                    LOGGER.info("Processing mappings for version", version=version.version_id, create_time=version.create_time)
 
                     try:
                         diff = _get_diff(base_repo, version.revision, prev_version.revision)
@@ -406,7 +406,7 @@ def _get_flipped_tasks_per_build(
     """
     try:
         prev_build: Build = prev_version.build_by_variant(build.build_variant)
-    except KeyError:
+    except (KeyError, AttributeError):
         LOGGER.warning(
             "Previous version does not contain a build for this build variant", exc_info=True
         )
@@ -414,7 +414,7 @@ def _get_flipped_tasks_per_build(
 
     try:
         next_build: Build = next_version.build_by_variant(build.build_variant)
-    except KeyError:
+    except (KeyError, AttributeError):
         LOGGER.warning(
             "Next version does not contain a build for this build variant", exc_info=True
         )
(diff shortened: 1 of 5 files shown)


src/selectedtests/task_mappings/create_task_mappings.py
```diff
@@ -406,15 +406,15 @@ def _get_flipped_tasks_per_build(
     """
     try:
         prev_build: Build = prev_version.build_by_variant(build.build_variant)
-    except KeyError:
+    except (KeyError, AttributeError):
```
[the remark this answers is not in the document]
jimoleary: :+1:

src/selectedtests/work_items/work_items_cli.py
```diff
@@ -82,16 +89,16 @@ def create_test_mapping(ctx: Context, project: str, src_regex: str, test_file_re
 
 @cli.command()
 @click.option(
-    "--weeks-back", type=int, default=DEFAULT_WEEKS_BACK, help="Number of weeks back to process."
+    "--years-back", type=int, default=DEFAULT_YEARS_BACK, help="Number of years back to process."
```
jimoleary: [suggestion] Values like weeks back and years back float around a bit and change over time (so when the years back was can be difficult to figure out later).

Why not just provide a start data and then leverage linux command line tools, e.g.:

```
$ date  --date="3 years ago" -I
2017-03-06

$ date  --date="24 weeks ago" -I
2019-09-20

$ work-items process-task-mappings  --start-date "$(date  --date="24 weeks ago" -I)"
$ work-items process-task-mappings  --start-date "$(date  --date="3 years ago" -I)"
```
 
This makes the CLI more flexible (and hope fully clear).

jimoleary (commented):
This is lgtm, but I have a suggestion about making the date range more flexible (and I think clear).

src/selectedtests/work_items/work_items_cli.py
lydiastepanek: Interesting, I didn't know about the `date` linux command, that's cool.

There are two concerns I have with making this change:
* Right now we are always running `work-items process-task-mappings` on a constant date range (and this provides consistency across the different projects we seed mappings for), so I don't want to make the CLI more flexible until it's needed.
* Occasionally I like to be able to run these scripts from my Mac when debugging. Adding Linux commands to the  `cronjobs/process_task_mapping_work_items.sh` script would make it only able to run on Linux. (We could add some logic to work on both Mac and Linux, but I don't think this is worth it right now.)

jimoleary: The date range would still be constant. 
The date cli should also work on a Mac.

It's only a suggestion so I have no problem ignoring it, we could raise a ticket to discuss it.

jimoleary (approved):

lydiastepanek:
evergreen merge

PR #48, opened 2020-03-09
Increase memory limit for process-task-mappings cronjob

lydiastepanek:


cronjobs.yml
```diff
@@ -15,7 +15,15 @@ cronJobs:
         memory: "5Mi"
       limits: # maximum allowed resources
         cpu: "1.5"
-        memory: "600Mi"
+        memory: "1.5Gi"
+    # kanopy recommends all cronjobs run on batch instance groups
```
lydiastepanek: @zamj pointed out that the Kanopy team now recommends that all cronjobs run on batch instance groups (and I confirmed with someone on Kanopy that they are updating their [docs](https://github.com/10gen/kanopy-docs/blob/master/docs/batch.md) to reflect this stance)

cronjobs.yml
```diff
@@ -15,7 +15,15 @@ cronJobs:
         memory: "5Mi"
       limits: # maximum allowed resources
         cpu: "1.5"
-        memory: "600Mi"
+        memory: "1.5Gi"
```
lydiastepanek: The process-test-mapping job is terminating due to an OOMKilled error so I'm increasing to the same limit we put on the process-task-mapping which is succeeding
Screen Shot 2020-03-09 at 3 19 10 PM

zamj (approved):
[nit] You can cut out some of the repetition if you want by using a yaml anchor for the node selector and toleration portions

lydiastepanek:
evergreen merge

PR #53, opened 2020-03-17
MAKE-1177: setup_ssh_keys.sh shouldn't overwrite keyfiles

ancostas:


cronjobs/lib/setup_ssh_keys.sh
```diff
@@ -8,15 +8,23 @@ setup_ssh_keys() {
...
-  chmod 600 $target_dir/id_rsa
+  chmod 600 $target_dir/selected_tests
 
-  cat > "$target_dir/id_rsa.pub" <<END_OF_PUBLIC
+  cat > "$target_dir/selected_tests.pub" <<END_OF_PUBLIC
 $GITHUB_PUBLIC_KEY
 END_OF_PUBLIC
+
+  eval "$(ssh-agent -s)"
+  ssh-add -k $target_dir/selected_tests
```
ancostas: I've done the ticket as recommended, but a couple things draw my attention.

Current API:
```
. cronjobs/lib/setup_ssh_keys.sh
setup_ssh_keys $target_dir 
```
* Will write to $target_dir/selected_tests + $target_dir/selected_tests.pub, failing if they already exist.

It seems like this could be simplified:
```
cronjobs/lib/setup_ssh_keys $target_file
```
* Will write to $target_file + $target_file.pub, failing if they already exist.
* Perhaps uses ~/.ssh/id_rsa as default $target_file?

jimoleary: I like the ```cronjobs/lib/setup_ssh_keys $target_file``` idea, it appears clearer to me. 

I also like the fact that it's not using id_ras now (I had to restore my copies of these files a little while ago).

The "don't  fail but just use what's there" was what I was trying to argue for at the planning meeting. But I was over-ruled. To be fair, these files don't really do a lot and are meant to be called from the container. So having it fail means that we will find issues quickly.

Would it be a good idea to check that GITHUB_PRIVATE_KEY and GITHUB_PUBLIC_KEY are not blank?

jimoleary (commented):
Overall I am good with the changes.

Simplifiying the script to take the filename seems a good idea to me.

jimoleary (approved):

ancostas:
evergreen merge

PR #55, opened 2020-03-27
MAKE-1205: Fix ssh key creation in cronjobs

dbradf:
Sourcing a file in bash does not let you pass an argument to it. Instead we can create a function that we can pass in an argument to.


jimoleary (approved):

dbradf:
evergreen merge

PR #57, opened 2020-04-13
Exploratory Friday: run integration tests for cronjobs in drone

lydiastepanek:


.drone.yml
```diff
@@ -85,3 +85,24 @@ pipeline:
     when:
       branch: master
       event: push
+
+  unit_tests:
+    image: 795250896452.dkr.ecr.us-east-1.amazonaws.com/server-tig/${DRONE_REPO_NAME}:latest
```
lydiastepanek, line 90: I asked the kanopy team and they said after the dronev1 upgrade we should be able to use the local Dockerfile to build the image, so this would change to:
```
image: .
```

.drone.yml
```diff
@@ -85,3 +85,24 @@ pipeline:
...
+    - pip3 install poetry
+    - poetry install
+    - poetry run black --check --diff src tests
+
+  check_mypy:
+    image: 795250896452.dkr.ecr.us-east-1.amazonaws.com/server-tig/${DRONE_REPO_NAME}:latest
+    commands:
+    - pip3 install poetry
+    - poetry install
+    - poetry run pytest --mypy src
```
ancostas, line 108: We could probably move this stuff into config files so its just `poetry run pytest` for all the tests.

src/selectedtests/helpers.py
```diff
@@ -28,7 +28,7 @@ def get_mongo_wrapper() -> MongoWrapper:
     """
     mongo_uri = os.environ.get("SELECTED_TESTS_MONGO_URI")
     if mongo_uri is None:
-        raise RuntimeError("Cannot connect to mongodb, SELECTED_TESTS_MONGO_URI is not set")
+        raise RuntimeError("Cannot connect to mongodb, SELECTED_TESTS_MONGO_URI is not set looooooooooong word")
```
ancostas, line 31: ?

ancostas (commented):
Probably not necessary, but I think it would be neat to spin up a pod w/ the API, a MDB with some seed data, and run an E2E test on the API. Is such a thing possible with Kanopy as it currently is?

src/selectedtests/helpers.py
ancostas: Nvm, you mentioned reducing the line length so this is to trip a failure I think.

.drone.yml
lydiastepanek: Sure, this is just an exploration PR though

src/selectedtests/helpers.py
lydiastepanek: see git commit message for explanation :p

.drone.yml
lydiastepanek: (This actually may take longer to roll out than I initially said in this comment ^)

.drone.yml
```diff
@@ -1,87 +1,156 @@
...
+  - set -xe apt-get install -y git
+  - pip3 install poetry
+  - poetry install
+  - sleep 5
+  - poetry run init-mongo --mongo-uri mongo_service create-indexes
+  when:
+    event:
+    - pull_request
+
+services:
```
lydiastepanek, line 152: Reference: https://docs.drone.io/pipeline/docker/syntax/services/

PR #58, opened 2020-04-24
Friday Exploration: Dependency Injection in Selected Tests

ancostas:
So, this seems nice!

Some benefits of this:
* Was able to add DI for REST API w/o needing to rewrite all the code or add it to the CLI (incremental updates/AGILE)
* REST API decoupled from data persistence layer

Also, we found a bug? in the `inject` package relating to the `__future__` module -- will open an issue with them regarding this.


src/selectedtests/app/app.py
```diff
@@ -8,14 +10,13 @@
     project_test_mappings_controller,
 )
 from selectedtests.datasource.mongo_wrapper import MongoWrapper
+from selectedtests.helpers import get_evg_api, get_mongo_wrapper
 
 
-def create_app(mongo_wrapper: MongoWrapper, evg_api: EvergreenApi) -> FastAPI:
+def create_app() -> FastAPI:
```
ancostas, line 16: These can be injected in tests using the container, rather than needing to be passed (a common theme for this DI stuff).

src/selectedtests/app/app.py
```diff
@@ -37,6 +38,12 @@ def create_app(mongo_wrapper: MongoWrapper, evg_api: EvergreenApi) -> FastAPI:
...
         tags=["projects"],
     )
-    app.state.db = mongo_wrapper
-    app.state.evg_api = evg_api
+
+    def dependencies(binder: inject.Binder):
+        binder.bind(EvergreenApi, get_evg_api())
+        binder.bind('EvergreenApi', get_evg_api())
+        binder.bind(MongoWrapper, get_mongo_wrapper())
+        binder.bind('MongoWrapper', get_mongo_wrapper())
```
ancostas, line 46: These string bindings are needed since `from __future__ import annotations` makes all of our type annotations strings instead --> `inject` breaks when looking for these strings since we only registered classes. Will talk with `inject` maintainers if it makes sense to register these automatically if omitted since otherwise their package doesn't mesh well with any classes that reference themselves.

src/selectedtests/app/controllers/project_task_mappings_controller.py
```diff
@@ -55,21 +52,17 @@ def get(
...
-    :param db: The database.
     :param project: The evergreen project.
     :param changed_files: List of source files to calculate correlated tasks for.
     :param threshold: Minimum threshold desired for flip_count / source_file_seen_count ratio
     """
-    evg_project = try_retrieve_evergreen_project(project, evg_api)
+    evg_project = try_retrieve_evergreen_project(project)
     task_mappings = get_correlated_task_mappings(
-        db.task_mappings(), parse_changed_files(changed_files), evg_project.identifier, threshold
+        parse_changed_files(changed_files), evg_project.identifier, threshold
```
ancostas, line 65: API just relies on the modules that do this stuff to do it without help, rather than needing to give them their own dependencies.

src/selectedtests/app/dependencies.py
```diff
@@ -1,26 +0,0 @@
...
-
-
-def get_evg(request: Request) -> EvergreenApi:
-    """
-    Get the configured Evergreen API client for the application.
-
-    :param request: The request needing the Evergreen API client.
-    :return: The Evergreen API client.
-    """
-    return request.app.state.evg_api
```
ancostas, line 26: I didn't much like FasstAPI's janky DI anyways...

src/selectedtests/app/evergreen.py
```diff
@@ -1,11 +1,13 @@
 """Web utilities for interacting with Evergreen API."""
+import inject
 
 from evergreen import EvergreenApi, Project
 from fastapi import HTTPException
 
 from selectedtests.evergreen_helper import get_evg_project
 
 
+@inject.autoparams()
```
ancostas, line 10: You'll see this in a few places; ideally it would be removed as well and we would maybe refactor this to be a bit more OOP oriented. With a refactor, you can just register injected constructors for each class, and they won't know about `inject` at all.

src/selectedtests/injection_helper.py
```diff
@@ -0,0 +1,22 @@
...
+        new_args = []
+        injector = inject.get_injector_or_die()
+        for arg_name, anno in arg_map.items():
+            if anno in injector._bindings:
+                new_args.append(injector.get_instance(anno))
+            else:
+                new_args.append(args.pop())
+        return func(*new_args, **kwargs)
+
+    return partial
```
ancostas, line 22: This is to avoid refactoring signatures like `func(dep, arg)` to `func(arg, dep)`. This will inject `dep` even if it comes before `arg`, if something is registered for that annotation.

src/selectedtests/task_mappings/get_task_mappings.py
```diff
@@ -2,21 +2,26 @@
...
     Retrieve task mappings associated with a given evergreen project and list of source files.
 
     :param project: The name of the evergreen project to analyze.
-    :param collection: Collection to act on.
+    :param db: db containing task mappings collection.
     :param changed_source_files: List of source files for which task mappings should be retrieved.
     :param threshold: Min threshold desired for flip_count/source_file_seen_count ratio.
     :return: A list of task mappings for the changed files.
     """
+    collection = db.task_mappings()
```
ancostas, line 24: This makes things easier. Arguably, searching for correlated task mappings in a collection other than `db.task_mappings()` doesn't really make sense anyway.

ancostas: Similar logic below for inserting things into/getting things from collections.

tests/selectedtests/work_items/test_task_mappings_work_item.py
```diff
@@ -27,46 +27,41 @@ def test_create_new_task_mappings(self):
...
 
-        collection.insert_one.assert_called_once()
+        mock_task_mappings_queue.insert_one.assert_called_once()
 
-        assert new_item
-
-    def test_insert_adds_item_to_collection(self):
-        collection = MagicMock()
-        collection.insert_one.return_value.acknowledged = True
+        assert new_item is True
```
ancostas, line 39: A little refactor, I think we should check that `True` is returned, rather than just a truthy value.

lydiastepanek (approved):
I'm not against this, so am approving. I don't think it makes the code noticeably easier to understand (especially since it adds more lines than it removes), but that could just be me.

src/selectedtests/app/app.py
```diff
@@ -37,6 +38,13 @@ def create_app(mongo_wrapper: MongoWrapper, evg_api: EvergreenApi) -> FastAPI:
         prefix="/projects/{project}/test-mappings",
         tags=["projects"],
     )
-    app.state.db = mongo_wrapper
-    app.state.evg_api = evg_api
+
+    def dependencies(binder: inject.Binder):
+        binder.bind(EvergreenApi, get_evg_api())
+        binder.bind("EvergreenApi", get_evg_api())
```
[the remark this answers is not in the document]
ancostas: This was the bug I ended up making a contribution to `python-inject` to fix. Basically, in order to do a self reference in a class a la:
```
class A:
    @classmethod
    def create_an_a() -> A:
        pass
```

You need to `from __future__ import annotations`. This converts all function annotations to strings, with the idea that they are meant as forward references to a class not yet defined, and allows mypy/the interpreter to delay their evaluation.

This clashes with string based annotation in `python-inject` -- in the version this was written with this registers the missing "forward references", but in the latest version this is fixed/not required.

src/selectedtests/app/app.py
```diff
@@ -37,6 +38,13 @@ def create_app(mongo_wrapper: MongoWrapper, evg_api: EvergreenApi) -> FastAPI:
         prefix="/projects/{project}/test-mappings",
         tags=["projects"],
     )
-    app.state.db = mongo_wrapper
-    app.state.evg_api = evg_api
+
+    def dependencies(binder: inject.Binder):
```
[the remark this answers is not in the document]
ancostas: So I'm wondering about that -- can we not just pass around the dependencies like normal in the *_cli.py scripts? One of the goals for the exercise was to only need to do DI in the app, and not have to fixup the cli scripts (migrating in pieces has planning/risk advantages).

I tried to get their tests working without using DI on them -- are there other cases I should ensure still work?

ancostas:
@lydiastepanek We don't need to merge this, was more an exercise, and probably needs more testing.

src/selectedtests/app/app.py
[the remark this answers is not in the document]
lydiastepanek: Ah, I see, thanks for explaining :)

src/selectedtests/app/app.py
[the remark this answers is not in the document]
lydiastepanek: RE migrating in pieces, I see. Yeah I think if we were going to switch to the `inject` library we should do it in pieces, that makes sense.

Since the web app tests pass, and you only changed the web app code, I don't think there are other pieces that need to be tested.

lydiastepanek:
@ancostas Oh okay cool. Also, just to clarify: We were already using DI in the app, right? My understanding was that this PR implements DI using the `inject` library specifically. Is that correct?

ancostas:
@lydiastepanek I wouldn't call it DI just yet, since we are manually assembling the software components (instantiating and wiring up dependencies ourselves, "by hand" so to say). I'd say the code is DI ready, and `python-inject` is just providing a concrete IoC container and assembler.

PR #59, opened 2020-04-30
MAKE-1227 Configure selected tests pytest to run all the correct tests and checks

jimoleary:


pyproject.toml
```diff
@@ -46,6 +46,8 @@ scipy = "^1.4.1"
 jupyter = "^1.0.0"
 matplotlib = "^3.1.3"
 pytest-cov = "^2.8.1"
+pytest-pydocstyle = "^2.1.3"
```
lydiastepanek, line 49: Are you able to remove `pydocstyle` (line 40) and `black` (line 38) as dev dependencies, or are they needed for `pytest-pydocstyle` and `pytest-black` to work?

evergreen.yml
```diff
@@ -37,7 +37,7 @@ tasks:
         set -o errexit
         . venv/bin/activate
 
-        poetry run pytest --flake8 --isort --junitxml=test_output_junit.xml
+        poetry run pytest --junitxml=test_output_junit.xml
```
lydiastepanek, line 40: I think you can remove the `check_formatting` and `check_pydoc` tasks (below), since they're duplicated by the `unit_tests` task now.

lydiastepanek (changes requested):

lydiastepanek (approved):

pyproject.toml
ancostas: FWIW @dbradf likes listing those out explicitly, rather than having them pulled in as transient dependencies (through pytest-pydocstyle/pytest-black).

ancostas: May be worth doing for sake of consistency.

ancostas (approved):

pyproject.toml
jimoleary: That's not what [shrub.py](https://github.com/evergreen-ci/shrub.py/blob/master/pyproject.toml#L20-L24) does.

But I'm happy to be consistent as long as I know what to be consistent with.

jimoleary: Since the dependant requirements are still pinned despite not being listed, I am going to leave it as is. It's easy to 'fix' if we decide that we should explicitly list requirements (but I would argue that is the job of the package owner).

